PR #21, opened 2014-11-18
Swift remote documentation

kahing:


andrewgaul:
Added https://github.com/andrewgaul/s3proxy/wiki/Provider-examples .

Issue #2, opened 2014-07-27
native multi-part upload


andrewgaul:
AWS-S3 limits single-part uploads to 5 GB and provides multi-part uploads to allow larger blobs.  This API consists of initiate, uploadPart, complete, abort, and listParts.  How to translate these multiple calls into the single jclouds method `BlobStore.putBlob(blob, multipart())`?  Can S3Proxy provide a dynamic `Payload` variant which demuxes these calls?

PR #26, opened 2015-01-03
Emulate multipart upload with single-part uploads

andrewgaul:
This approach requires three times as many operations as the optimal
approach.  Implementing this correctly requires exposing the
underlying multipart operations in jclouds.  Most s3-tests pass but
two still fail:

test_multipart_upload_size_too_small
test_list_multipart_upload

References #2.


andrewgaul:
This implements the multipart API at a ruinous cost -- writing, reading, and re-writing every part.  I am not sure if it is worth merging but it does shake out all the required functionality that we need to migrate into jclouds.

PR #43, opened 2015-02-22
Skip install goal due to MSHADE-183

andrewgaul:


andrewgaul:
This does not work since `run-s3-tests.sh` needs the s3proxy artifact.

PR #47, opened 2015-03-06
Add a test for failure when using v4 signatures.

timuralp:
Adds a unit test with the Amazon Java SDK for failures when submitting
v4-signed requests.

### DIFF
--- a/src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
+++ b/src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright 2014 Andrew Gaul <andrew@gaul.org>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Test
+    public void testAwsV4Failure() throws Exception {
+        BasicAWSCredentials awsCreds = new BasicAWSCredentials(s3Identity,
+                s3Credential);
+        AmazonS3 client = new AmazonS3Client(awsCreds);
+        client.setEndpoint(s3Endpoint.toString());
+        try {
+            client.getObject(containerName, "foo");
```
andrewgaul: You need to call `fail()` call here to ensure that `S3Proxy` throws the expected exception.

timuralp:
Updated to use the ExpectedException rule.

src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
```diff
@@ -0,0 +1,172 @@
...
+            s3Context.close();
+        }
+        if (context != null) {
+            context.getBlobStore().deleteContainer(containerName);
+            context.close();
+        }
+    }
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
```
andrewgaul: If you rename this to `thrown` it will ignore the visibility.

andrewgaul:
Commit message should mention that we use the AWS SDK instead of jclouds due to [JCLOUDS-480](https://issues.apache.org/jira/browse/JCLOUDS-480).

andrewgaul:
Please address Checkstyle nits then we can merge this.

PR #53, opened 2015-04-28
handle empty prefix as if it's missing

kahing:


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -791,7 +791,7 @@ private void handleBlobList(HttpServletRequest request,
             options = options.recursive();
         }
         String prefix = request.getParameter("prefix");
-        if (prefix != null) {
+        if (prefix != null && !prefix.isEmpty()) {
```
andrewgaul: Could call `!Strings.isNullOrEmpty()`

andrewgaul:
:+1:

PR #54, opened 2015-04-29
parse authorization header backwards

kahing:
this is to accomodate using s3proxy with swift backend, which has
a colon in the access key


andrewgaul:
:+1:

PR #59, opened 2015-05-08
get rid of default blob store

kahing:


andrewgaul:
Pushed to master as 16bb2082f4c25a61a54c9b8d3d032e34fa850711.

PR #60, opened 2015-05-12
correctly follow the host header documentation

kahing:
https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#VirtualHostingSpecifyBucket

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -223,10 +223,15 @@ final class S3ProxyHandler extends AbstractHandler {
         if (hostHeader != null && virtualHost.isPresent()) {
             hostHeader = HostAndPort.fromString(hostHeader).getHostText();
             String virtualHostSuffix = "." + virtualHost.get();
-            if (hostHeader.endsWith(virtualHostSuffix)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -223,10 +223,15 @@ private void doHandle(HttpServletRequest request,
         if (hostHeader != null && virtualHost.isPresent()) {
             hostHeader = HostAndPort.fromString(hostHeader).getHostText();
             String virtualHostSuffix = "." + virtualHost.get();
-            if (hostHeader.endsWith(virtualHostSuffix)) {
-                String bucket = hostHeader.substring(0,
-                        hostHeader.length() - virtualHostSuffix.length());
-                uri = "/" + bucket + uri;
+            if (!hostHeader.equals(virtualHost.get())) {
```
andrewgaul: Push this check into the conditional above?

andrewgaul:
:+1:

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
kahing: can't because hostHeader is initialized inside the outer if

PR #61, opened 2015-05-12
add a way to listen to both http and https

kahing:


src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -57,10 +57,12 @@ public static void main(String[] args) throws Exception {
...
+        String secureEndpoint = properties.getProperty(
+                S3ProxyConstants.PROPERTY_SECURE_ENDPOINT);
         String s3ProxyAuthorization = properties.getProperty(
                 S3ProxyConstants.PROPERTY_AUTHORIZATION);
         if (provider == null || identity == null || credential == null ||
-                s3ProxyEndpointString == null ||
+                (s3ProxyEndpointString == null && secureEndpoint == null) ||
                 s3ProxyAuthorization == null) {
             System.err.println("Properties file must contain:\n" +
                     Constants.PROPERTY_PROVIDER + "\n" +
```
andrewgaul: Update error message?

src/main/java/org/gaul/s3proxy/S3Proxy.java
```diff
@@ -144,6 +162,10 @@ public int getPort() {
         return ((ServerConnector) server.getConnectors()[0]).getLocalPort();
     }
 
+    public int getSecurePort() {
+        return ((ServerConnector) server.getConnectors()[1]).getLocalPort();
+    }
+
```
andrewgaul: Does this throw `ArrayOutOfBoundException` when a user only configured HTTPS?

andrewgaul:
Why do the integration tests fail?

kahing:
does travis generate the keys in keystore? it's complaining that keystore doesn't exist

andrewgaul:
`run-s3-tests.sh` only configures HTTP; can we just disable the secure endpoint for this test?

andrewgaul:
Rebased and pushed to master as 6e4ebad1c8982cc0259fbdafedb4434ef6955119, f17d968a3341a80a7d80e12aa56b16f76d3c4604, and 170aedb319cc1b9b692d00b93d1a946cb876586a.

PR #67, opened 2015-06-24
jclouds integration test

kahing:


andrewgaul:
I ran these tests 100 times with both JDK 1.7 and 1.8 and cannot reproduce the Travis symptoms

kahing:
the maven tests passed

PR #72, opened 2015-07-17
Partially emulate arbitrary markers with Azure

andrewgaul:


andrewgaul:
@kahing Please review.

PR #78, opened 2015-08-22
Terminate on s3proxy start failure

flandr:
E.g. when bind(2) fails.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/Main.java
+++ b/src/main/java/org/gaul/s3proxy/Main.java
@@ -131,6 +131,11 @@ public final class Main {
             System.exit(1);
             throw e;
         }
-        s3Proxy.start();
+        try {
+            s3Proxy.start();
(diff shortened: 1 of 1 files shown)


src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -131,6 +131,10 @@ public static void main(String[] args) throws Exception {
             System.exit(1);
             throw e;
         }
-        s3Proxy.start();
+        try {
+            s3Proxy.start();
+        } catch (Exception e) {
+            System.exit(1);
```
andrewgaul: Log exception or at least the message?

flandr: Dumped it to stderr

andrewgaul:
Thank you for your contribution @flandr!

Issue #24, opened 2014-11-19
AWS signature V4


andrewgaul:
Amazon allows V2 signatures for old regions and requires [V4 signatures](http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) for new regions like eu-central-1 (Frankfurt).  jclouds plans to move to V4 by default for aws-s3 [JCLOUDS-480](https://issues.apache.org/jira/browse/JCLOUDS-480) so we must support V4 to allow jclouds applications to use S3Proxy.

andrewgaul:
Pull request for tests: ceph/s3-tests#84

andrewgaul:
Work in progress commit: https://github.com/timuralp/s3proxy/commit/a80b1c94245d07550a55bcd50a7fe93100989fa2

PR #84, opened 2015-10-11
Do not check authorization when in anonymous mode

andrewgaul:


andrewgaul:
@kahing This should allow use of aws-go-sdk until it implements v2 signing or S3Proxy implements v4 signing in #24.

kahing:
I used Credentials: credentials.AnonymousCredentials to workaround the credential problem, and hit the regions problem next (aws-sdk-go cannot omit region). Doesn't look like this PR addresses that?

PR #86, opened 2015-10-16
Handle listing a public-read container

andrewgaul:
Also handle checking existence.


andrewgaul:
Tests coming in ceph/s3-tests#83.

PR #95, opened 2015-11-17
fix(S3ProxyHandler) - Anonymous access mode should return all buckets…

### LINKED ISSUE #94: Anonymous access mode cannot list a private access bucket
Anonymous access configuration should skip `doHandleAnonymous` handling which only allows access to public-read buckets and keys.  Discovered by @ritazh and @sedouard.

ritazh:
Tested against S3 and Azure Blob Storage
Closes #94


andrewgaul:
Rebased, reworded commit message, and pushed to master as 71e2d6137781db6d047ee43e8f7b699539937a71.  Thank you for your contribution @ritazh!

Issue #100, opened 2015-12-02
Multipart upload fails when multiple UploadPartRequests are needed


ritazh:
With multiple UploadPartRequest, only the first UploadPartRequest actually works. All subsequent UploadPartRequests are processed but blob is not actually uploaded. It's possible that I may have missed something.

My test is with a file of size 17MB, part size of 10MB. The expected behavior should be after CompleteMultipartUploadRequest, the entire 17MB file should be uploaded. The actual is a blob of size a little less than 10MB was uploaded as a result from the first UploadPartRequest. Yet a new ETag was returned for both UploadPartResult. From S3Proxy log, seems both requests for partnumber 1 and 2 have been processed properly.

Here is an updated testBigMultipartUpload PR #101 that tests the same behavior. Note the test currently fails.

Log for your reference: [Full Log here](https://gist.github.com/ritazh/bc29fead323e3a527e61)
After the first PUT request, we see binaries in the log:
```
request: (PUT /MyObjectKeyslack?uploadId=t1ZKU.ywRVW7J0ce5asSJN0ZbaiqfLE_A0ovM7gsPEmNtu87bhvdK255Gfw43aTnnUd9lspldyvKbu0u4FX7lPVcE351yGgF5_4OE6gzi.s-&partNumber=1)@265700430 org.eclipse.jetty.server.Request@fd6444e
```
After the second PUT request, no binaries seem to be uploaded in the log.
```
request: (PUT /MyObjectKeyslack?uploadId=t1ZKU.ywRVW7J0ce5asSJN0ZbaiqfLE_A0ovM7gsPEmNtu87bhvdK255Gfw43aTnnUd9lspldyvKbu0u4FX7lPVcE351yGgF5_4OE6gzi.s-&partNumber=2)@265700430 org.eclipse.jetty.server.Request@fd6444e
```

andrewgaul:
Now that you corrected the test and Travis passes, is this still a valid issue?

ritazh:
Yes. Thanks for your help. I'm able to upload the entire file now. Hope the test case will help others in the future.

PR #101, opened 2015-12-02
fix(S3AwsSdkTest) - update testBigMultipartUpload

ritazh:
Test multipart upload when upload blob size is greater than partSize, resulting in multiple UploadPartRequests.

This test emulate the same behavior as #100 

NOTE: This test currently fails.
```
Results :
Failed tests: 
  S3AwsSdkTest.testBigMultipartUpload:275 
InputStreams do not have equal content
```

### DIFF
--- a/src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
+++ b/src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
@@ -219,7 +219,10 @@ public final class S3AwsSdkTest {
         client.setEndpoint(s3Endpoint.toString());
 
         String key = "multipart-upload";
-        int size = 10_000_000;
+        // If size <= partSize, then only one UploadPartRequest is needed
+        // to upload all of the blob content.
+        int size = 9_000_000;
+        int partSize = 5_000_000;
         ByteSource byteSource = TestUtils.randomByteSource().slice(0, size);
 
         InitiateMultipartUploadRequest initRequest =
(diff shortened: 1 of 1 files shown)


src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
```diff
@@ -238,13 +241,24 @@ public void testBigMultipartUpload() throws Exception {
...
 
         UploadPartResult uploadPartResult = client.uploadPart(uploadRequest);
         PartETag partETag = uploadPartResult.getPartETag();
+
+        // Second uploadRequest with partNumber = 2 to upload rest of the blob
+        UploadPartRequest uploadRequest2 = new UploadPartRequest()
+                .withBucketName(containerName).withKey(key)
+                .withUploadId(uploadId).withPartNumber(2)
+                .withInputStream(byteSource.openStream())
+                .withObjectMetadata(metadata);
```
andrewgaul: This test is incorrect.  It succeeds when adding:

```java
.withInputStream(byteSource.slice(partSize, partSize).openStream())
.withPartSize(partSize);
```

ritazh: You are right. Updated this test with the suggested changes.

andrewgaul:
Fixed the part size to match the AWS minimum size, reverted some unintentional changes, reworded commit message, and pushed to master as a5221f5f8e51c639c958dfca5fa9799b572d1803.

Issue #96, opened 2015-11-18
AWS SDK upload fails with Azure backend


ritazh:
Using this sample code here:
https://github.com/aws/aws-sdk-java/blob/master/src/samples/AmazonS3/S3Sample.java#L119

Error:
at com.amazonaws.services.s3.AmazonS3Client.putObject(AmazonS3Client.java:1398): Input is expected to be encoded in multiple of 2 bytes but found: 17

File was uploaded successfully, but ends in exception.

Same test runs successfully against AWS S3.

I see this PR added etag support for AzureBlob. https://github.com/jclouds/jclouds/pull/716/files

ritazh:
Adding more info, in `com.amazonaws.util`, `Base16Codec` is expecting an even length.

```
public byte[] decode(byte[] src, final int length)
    {
        if (length % 2 != 0) {
            throw new IllegalArgumentException(
                "Input is expected to be encoded in multiple of 2 bytes but found: "
                + length
            );
        }
        final byte[] dest = new byte[length / 2];
```

kahing:
s3 expects a regular PUT to return etags as the md5sum of the blob encoded in base16, and the `AmazonS3Client` is probably verifying the md5sum from the response. Maybe S3Proxy can return Content-MD5 as etags instead?

ritazh:
Yes I think that would work.

kahing:
@andrewgaul pointed out to me that munging with etags will cause problems with conditional GETs. And the option in the sdk to disable the check is not implemented correctly: https://github.com/aws/aws-sdk-java/blob/bb9e0be441afb5658f0edbbd19bcf27e2fac4105/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServiceUtils.java#L452

ritazh:
I'm trying to figure out why jcloud returns an ETag of 17 bytes for Azure blob. Any suggestions?

kahing:
According to https://github.com/jclouds/jclouds/pull/716/files#diff-e724f00ae88c3a1a9ec1ddbe252a3bccL10 Azure etags looks like 0x8CACB9BD7C1EEEC which is 17 bytes.

andrewgaul:
[...]

PR #102, opened 2015-12-08
For Azure backend, map Content-MD5 to  ETag for PUT blob

ritazh:
This fixes issue #96

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -425,6 +425,9 @@ final class S3ProxyHandler extends AbstractHandler {
             if (headerName.startsWith("x-amz-meta-")) {
                 continue;
             }
+            if (headerName.startsWith("x-amz-content-")) {
+                continue;
+            }
             if (!SUPPORTED_X_AMZ_HEADERS.contains(headerName)) {
                 logger.error("Unknown header {} with URI {}",
                         headerName, request.getRequestURI());
@@ -1386,6 +1389,10 @@ final class S3ProxyHandler extends AbstractHandler {
                 }
             }
 
+            if (blobStoreType.equals("azureblob")) {
+                eTag = BaseEncoding.base16().lowerCase()
+                        .encode(BaseEncoding.base64().decode(contentMD5String));
+            }
             response.addHeader(HttpHeaders.ETAG, maybeQuoteETag(eTag));
         }
 


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -1386,6 +1389,10 @@ private void handlePutBlob(HttpServletRequest request,
                 }
             }
 
+            if (blobStoreType.equals("azureblob")) {
+                eTag = BaseEncoding.base16().lowerCase()
+                        .encode(BaseEncoding.base64().decode(contentMD5String));
+            }
```
andrewgaul: This returns the (possibly-null) MD5 client header back to the client.  Instead we should return the MD5 header from the server, allowing the client to do end-to-end verification of uploaded data.

andrewgaul:
This pull request is incomplete; the commit message mentions GET, HEAD, and list but only modifies PUT.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -425,6 +425,9 @@ private void doHandle(HttpServletRequest request,
             if (headerName.startsWith("x-amz-meta-")) {
                 continue;
             }
+            if (headerName.startsWith("x-amz-content-")) {
+                continue;
+            }
```
andrewgaul: Spurious change?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
ritazh: Thanks for the feedback. I could be wrong about this but the AWS Java SDK expects a base16 encoded ETag header. If server returns an ETag like #96 , Client app will fail. While MD5 header is returned to the client for PUT, unless client app does extra work to validate, the current Java SDK will fail with invalid value returned for ETag. I can definitely add checks for null MD5 client header.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
ritazh: No. For Azure backend, this caused an error `Unknown header x-amz-content-sha256 with URI`. Sorry I should not have mixed this here. Let me know if I should create a different PR for this or another way to fix it.

ritazh:
Updated the commit comment to just PUT blob since GET, LIST, all trigger HEAD request and HEAD request does not return MD5 in the header.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
andrewgaul: Let's ignore base16 and base64 encoding for a moment.  The client may provide a Content-MD5 header during a PUT which the server uses to check end-to-end data integrity.  During this operation AWS always returns an ETag which is a MD5 hash of the object.  Clients can use this to ensure the end-to-end integrity of the PUT request.  This pull request removes this feature by returning the *client* Content-MD5 instead of the *server* ETag MD5.  Even if you add a null check for the client Content-MD5, what value will you return for ETag instead?  This pull request also removes support for conditional GET by changing the ETag.

I do not want to discourage work on a higher-fidelity mapping of S3 to Azure, but have you thought through other issues like multi-part upload which returns yet another form of ETag?  Doing a complete emulation of Content-MD5 ETag will be quite involved including mapping the GET, HEAD, and list operations.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
andrewgaul: Sorry I don't understand; can you show me a test case that this fails with?  Let's use a separate pull request for the test and fix.

PR #106, opened 2015-12-11
initial support for encoding-type

dleute:
This should enable support. I think I caught all the places blobName is put into xml data.

Take a look!

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -18,12 +18,7 @@ package org.gaul.s3proxy;
 
 import static java.util.Objects.requireNonNull;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.PushbackInputStream;
-import java.io.Writer;
+import java.io.*;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
@@ -141,6 +136,7 @@ final class S3ProxyHandler extends AbstractHandler {
             "AWSAccessKeyId",
             "delete",
             "delimiter",
+            "encoding-type",
             "Expires",
             "location",
             "marker",
@@ -149,8 +145,7 @@ final class S3ProxyHandler extends AbstractHandler {
             "prefix",
             "Signature",
             "uploadId",
-            "uploads",
-            "encoding-type"
+            "uploads"
     );
     /** All supported x-amz- headers, except for x-amz-meta- user metadata. */
     private static final Set<String> SUPPORTED_X_AMZ_HEADERS = ImmutableSet.of(
@@ -2283,8 +2278,12 @@ final class S3ProxyHandler extends AbstractHandler {
 
     // Protects against illegal characters in blob names only if the client requests it
     private String encodeBlob(String encodingType, String blobName) {
-        if (encodingType.equals("url")) {
-            return URLEncoder.encode(blobName, "UTF-8");
+        if (encodingType != null && encodingType.equals("url")) {
+            try {
+                return URLEncoder.encode(blobName, "UTF-8");
+            } catch (UnsupportedEncodingException e) {
+                return blobName;
+            }
         } else {
             return blobName;
         }
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -955,6 +955,7 @@ final class S3ProxyHandler extends AbstractHandler {
         String blobStoreType = getBlobStoreType(blobStore);
         ListContainerOptions options = new ListContainerOptions();
         String delimiter = request.getParameter("delimiter");
+        String encodingType = request.getParameter("encoding-type");
         if (delimiter != null) {
             options.delimiter(delimiter);
         } else {
@@ -1002,7 +1003,8 @@ final class S3ProxyHandler extends AbstractHandler {
             if (prefix == null) {
                 xml.writeEmptyElement("Prefix");
             } else {
-                writeSimpleElement(xml, "Prefix", prefix);
+                writeSimpleElement(xml, "Prefix", encodeBlob(
+                        request.getParameter("encoding-type"), prefix));
             }
 
             writeSimpleElement(xml, "MaxKeys", String.valueOf(maxKeys));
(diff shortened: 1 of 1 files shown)


dleute:
This should fix issue #106.

I have not figured out the best way to test this yet.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -148,7 +149,8 @@
             "prefix",
             "Signature",
             "uploadId",
-            "uploads"
+            "uploads",
+            "encoding-type"
```
andrewgaul: Can you alphabetize these?

andrewgaul:
This fails Travis, can you look at the compilation errors (and likely Checkstyle violations)?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2280,4 +2280,13 @@ private static String maybeQuoteETag(String eTag) {
     private static boolean startsWithIgnoreCase(String string, String prefix) {
         return string.toLowerCase().startsWith(prefix.toLowerCase());
     }
+
+    // Protects against illegal characters in blob names only if the client requests it
+    private String encodeBlob(String encodingType, String blobName) {
+        if (encodingType.equals("url")) {
```
andrewgaul: `HttpServletRequest.getParameter` can return null and thus we must check `encodingType` here.

dleute: .equals in the function does that by default. Supposedly.

Sent from my iPhone

On Dec 11, 2015, at 9:25 PM, Andrew Gaul <notifications@github.com> wrote:

In src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
<https://github.com/andrewgaul/s3proxy/pull/106#discussion_r47425453>:

> @@ -2280,4 +2280,13 @@ private static String maybeQuoteETag(String eTag) {
>      private static boolean startsWithIgnoreCase(String string, String prefix) {
>          return string.toLowerCase().startsWith(prefix.toLowerCase());
>      }
> +
> +    // Protects against illegal characters in blob names only if the client requests it
> +    private String encodeBlob(String encodingType, String blobName) {
> +        if (encodingType.equals("url")) {

HttpServletRequest.getParameter can return null and thus we must check
encodingType here.

—
Reply to this email directly or view it on GitHub
<https://github.com/andrewgaul/s3proxy/pull/106/files#r47425453>.

andrewgaul:
Neither the AWS Java SDK nor jclouds support the url encoding-type presently so we cannot add a unit test.  Could you share a small aws-cli example so I can verify this manually?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
andrewgaul: `Object.equals` checks its parameters against null but if `encodingType` is null then we throw a `NullPointerException` before calling `equals`.

dleute:
With the lates awscli 1.9.12 just do an ls of a bucket. It always sends
encoding type for me.

Sent from my iPhone

On Dec 11, 2015, at 9:27 PM, Andrew Gaul <notifications@github.com> wrote:

Neither the AWS Java SDK nor jclouds support the url encoding-type
presently so we cannot add a unit test. Could you share a small aws-cli
example so I can verify this manually?

—
Reply to this email directly or view it on GitHub
<https://github.com/andrewgaul/s3proxy/pull/106#issuecomment-164101413>.

dleute:
I'll be working on this again on monday.

dleute:
FYI: swift has entirely spoiled me. I don't think about null pointer exceptions anymore.

dleute:
I just pushed in the requested changes. Two comments:

1. I handled the null pointer case in the function since that's where the actual danger is. Not sure if this fits your coding style. I felt it's cleaner than having the same check capability in many parts of the code.

2. I handled the exception, but I'm not sure what you would want the behavior when an unknown encoding is used. I just returned the original string so nothing crashes.It's using only utf-8 so this error should really never happen.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2280,4 +2275,17 @@ private static String maybeQuoteETag(String eTag) {
...
         return string.toLowerCase().startsWith(prefix.toLowerCase());
     }
+
+    // Protects against illegal characters in blob names only if the client requests it
+    private String encodeBlob(String encodingType, String blobName) {
+        if (encodingType != null && encodingType.equals("url")) {
+            try {
+                return URLEncoder.encode(blobName, "UTF-8");
+            } catch (UnsupportedEncodingException e) {
+                return blobName;
```
andrewgaul: Can you `throw Throwables.propagate(e)` since this exception should never be thrown?

andrewgaul:
Please address Checkstyle violations listed by Travis.

andrewgaul:
Looking at https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html it seems like we should handle `encoding-type` in list objects as well as emit the Encoding Type back to the client.

dleute:
Yeah. We also are supposed to encode those other things: ```Delimiter, Marker, Prefix, NextMarker, Key.```

Right now all we are doing is Key for blobNames.

I imagine we also only include and encode if the encoding-type parameter is "url".

dleute:
Just pushed in the updates based on that document. I tried to match the styles in the function already.

Take a look.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -1654,7 +1667,8 @@ private void handleListParts(HttpServletRequest request,
             xml.writeDefaultNamespace(AWS_XMLNS);
 
             writeSimpleElement(xml, "Bucket", containerName);
-            writeSimpleElement(xml, "Key", blobName);
+            writeSimpleElement(xml, "Key", encodeBlob(
+                    request.getParameter("encoding-type"), blobName));
```
andrewgaul: `handleListParts` needs an `Encoding-Type` element: https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadListParts.html

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2280,4 +2294,18 @@ private static String maybeQuoteETag(String eTag) {
     private static boolean startsWithIgnoreCase(String string, String prefix) {
         return string.toLowerCase().startsWith(prefix.toLowerCase());
     }
+
+    // Protects against illegal characters in
+    // blob names only if the client requests it
+    private String encodeBlob(String encodingType, String blobName) {
```
andrewgaul: Should be `static`.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
dleute: According to that link, the ancestor is ListBucketResult not ListPartsResult. Not sure if that's a typo, but that isn't in the ListPartsResult area. If you want I can add it and assume it's a typo in the documentation. See Ancestor in attached screenshot. (The ancestor of all the other items is ListPartsResult)

screen shot 2015-12-14 at 4 48 28 pm

andrewgaul:
I looked through the boto source which aws cli uses and it sends `encoding-type` for

* `list` and `get_all_keys`
* `list_versions`
* `list_multipart_uploads` and `get_all_multipart_uploads`
* `get_all_parts`

Based on this, I removed `encoding-type` handling from `handleMultiBlobRemove`, `handleInitiateMultipartUpload`, and `handleCompleteMultipartUpload`.  While I agree that your pull request seems more correct, I would rather err on the side of compatibility with boto which is one of the most popular clients.  I amended your commit, reworded the commit message, and pushed to master as d8c73bc86d30e885e743eb7afcd41cf0b9e9497b.  Thank you for your contribution @dleute!

dleute:
Happy to help!

Sent from my iPhone

On Dec 14, 2015, at 8:13 PM, Andrew Gaul <notifications@github.com> wrote:

I looked through the boto source which aws cli uses and it sends
encoding-type for

   - list and get_all_keys
   - list_versions
   - list_multipart_uploads and get_all_multipart_uploads
   - get_all_parts

Based on this, I removed encoding-type handling from handleMultiBlobRemove,
handleInitiateMultipartUpload, and handleCompleteMultipartUpload. While I
agree that your pull request seems more correct, I would rather err on the
side of compatibility with boto which is one of the most popular clients. I
amended your commit, reworded the commit message, and pushed to master as
d8c73bc
<https://github.com/andrewgaul/s3proxy/commit/d8c73bc86d30e885e743eb7afcd41cf0b9e9497b>.
Thank you for your contribution @dleute <https://github.com/dleute>!

—
Reply to this email directly or view it on GitHub
<https://github.com/andrewgaul/s3proxy/pull/106#issuecomment-164610343>.

PR #112, opened 2016-01-07
Support AWS v4 signatures

### LINKED ISSUE #24: AWS signature V4

andrewgaul:
Fixes #24.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/ChunkedInputStream.java
+++ b/src/main/java/org/gaul/s3proxy/ChunkedInputStream.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2014-2015 Andrew Gaul <andrew@gaul.org>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


andrewgaul:
@timuralp Please review.

src/main/java/org/gaul/s3proxy/ChunkedInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+            if (line.equals("")) {
+                return -1;
+            }
+            String[] parts = line.split(";", 2);
+            currentLength = Integer.parseInt(parts[0]);
+            currentSignature = parts[1];
+            chunk = new byte[currentLength];
+            currentIndex = 0;
+            ByteStreams.readFully(in, chunk);
+            // TODO: check chunkSinature
```
timuralp: chunkSignature

andrewgaul: Done.

timuralp:
This looks fine to me. Too bad the s3-tests project doesn't have v4 signer tests, but hopefully they'll move faster on that than they have on other issues. The SDK tests should help in the meantime.

andrewgaul:
s3-tests does have tests although I cannot run them as discussed in https://github.com/ceph/s3-tests/pull/84#issuecomment-170142987.

PR #114, opened 2016-01-14
Use uploadId for stub blob name

### LINKED ISSUE #108: Multipart upload not atomic
MPU starts by writing a stub object with the same name as the mpu key name, which overwrites the original object (if it existed) even if complete multipart upload is never submitted.

andrewgaul:
This makes MPU overwrites of an existing blob atomic.  Fixes #108.


andrewgaul:
@kahing Please review.

PR #125, opened 2016-02-20
Add EventualBlobStore

### LINKED ISSUE #65: Add mechanism to model eventual consistency
S3Proxy offers an opportunity to implement an eventual consistency middleware.  This would allow a more deterministic and reliable way to uncover applications erroneously relying on strong consistency.  This middleware could use two backend blobstores, writing the the first one and reading from the second, periodically syncing updates from the former to the latter.

andrewgaul:
This models eventually-consistent behavior.  This implementation uses
two buckets and with client writes going to the first bucket and reads
to the second.  Operations are replicated from the first to the second
with a variable delays.  A more complete implementation could flap
between the strongly- and eventually-consistent buckets and arbitrary
reorder operations which conclude with last-writer wins.  Fixes #65.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/EventualBlobStore.java
+++ b/src/main/java/org/gaul/s3proxy/EventualBlobStore.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright 2014-2016 Andrew Gaul <andrew@gaul.org>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gaul.s3proxy;
+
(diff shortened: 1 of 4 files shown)


andrewgaul:
@kahing Please review.

src/main/java/org/gaul/s3proxy/EventualBlobStore.java
```diff
@@ -0,0 +1,208 @@
...
+        MultipartUpload mpu = nearStore.initiateMultipartUpload(container,
+                blobMetadata, options);
+        delegate().initiateMultipartUpload(container, blobMetadata, options);
+        return mpu;
+    }
+
+    @Override
+    public void abortMultipartUpload(MultipartUpload mpu) {
+        nearStore.abortMultipartUpload(mpu);
+        delegate().abortMultipartUpload(mpu);
```
andrewgaul: This does not support non-repeatable `Payload`, e.g., `InputStream`.

src/main/java/org/gaul/s3proxy/EventualBlobStore.java
```diff
@@ -0,0 +1,208 @@
...
+                1 + random.nextInt(delay), delayUnit);
+        return nearETag;
+    }
+
+    @Override
+    public MultipartPart uploadMultipartPart(MultipartUpload mpu,
+            int partNumber, Payload payload) {
+        MultipartPart part = nearStore.uploadMultipartPart(mpu, partNumber,
+                payload);
+        delegate().uploadMultipartPart(mpu, partNumber, payload);
```
andrewgaul: This does not support non-repeatable `Payload`, e.g., `InputStream`.

PR #132, opened 2016-03-15
Add NullBlobStore

### LINKED ISSUE #131: Add middleware to test extremely large files
Some clients want to test extremely large objects, e.g., 10,000 part multi-part uploads.  Using the file system provider with the default S3 minimum 5 MB part size makes this prohibitive.  Instead S3Proxy could support a /dev/null mode which discards client PUTs and returns some deterministic result for GETs.

andrewgaul:
This throws away all data during object creation and reconstructs it
with NUL bytes.  Clients can use this to test very large objects.
Fixes #131.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/Main.java
+++ b/src/main/java/org/gaul/s3proxy/Main.java
@@ -160,6 +160,13 @@ public final class Main {
                     delay, TimeUnit.SECONDS, probability);
         }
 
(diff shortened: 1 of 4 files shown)


andrewgaul:
@kahing Please review.

andrewgaul:
Needs jclouds/jclouds#1037.

PR #123, opened 2016-02-17
Provide Dockerfile

andrewgaul:
Fixes #120.

PR #141, opened 2016-06-15
support for S3 response header overrides

zvikagart:
I added support for `response-content-disposition` request query parameter in object GETs.
See details: https://forums.aws.amazon.com/ann.jspa?annID=884

I also added a simple Dockerfile to support a docker development environment based on the official [maven docker image](https://hub.docker.com/_/maven/). You can then run `docker build -t s3proxy-build -f Dockerfile.build .` to build and test s3proxy. This encapsulates java and maven dependencies quite nicely.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -156,7 +156,13 @@ final class S3ProxyHandler extends AbstractHandler {
             "prefix",
             "Signature",
             "uploadId",
-            "uploads"
+            "uploads",
+            "response-cache-control",
+            "response-content-disposition",
+            "response-content-encoding",
+            "response-content-language",
+            "response-content-type",
+            "response-expires"
     );
     /** All supported x-amz- headers, except for x-amz-meta- user metadata. */
     private static final Set<String> SUPPORTED_X_AMZ_HEADERS = ImmutableSet.of(
@@ -1381,7 +1387,7 @@ final class S3ProxyHandler extends AbstractHandler {
         }
 
         response.setStatus(HttpServletResponse.SC_OK);
-        addMetadataToResponse(response, metadata);
+        addMetadataToResponse(request, response, metadata);
     }
 
     private void handleGetBlob(HttpServletRequest request,
@@ -1442,7 +1448,7 @@ final class S3ProxyHandler extends AbstractHandler {
 
         response.setStatus(status);
 
-        addMetadataToResponse(response, blob.getMetadata());
+        addMetadataToResponse(request, response, blob.getMetadata());
         // TODO: handles only a single range due to jclouds limitations
         Collection<String> contentRanges =
                 blob.getAllHeaders().get(HttpHeaders.CONTENT_RANGE);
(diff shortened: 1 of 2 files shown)


andrewgaul:
@zvikagart Can you add or modify a test for the `response-content-disposition` parameter?  Also please remove the Dockerfile from this pull request and can you review the more configurable one in #123?

zvikagart:
I tried to dig a little bit to find a way to test this query parameter, but it looks like jclouds doesn't support this yet (https://issues.apache.org/jira/browse/JCLOUDS-1016). If you have any pointers I'll gladly implement this.

As for the Dockerfile I added, I intentionally named it differently (`Dockefile.build`) since it serves a totally different role than the Dockerfile in #123. The one I added is just for building the project, not to run the service.

andrewgaul:
Could you use the AWS SDK instead of jclouds?  I believe you just need to add your header to `SignableRequest`:

https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/SignableRequest.html

You can find other tests using the AWS SDK in `S3AwsSdkTest`.

andrewgaul:
@zvikagart I added a Dockerfile so please add the test and rebase this PR so I can merge it!

zvikagart:
@andrewgaul, I generalized the code to support the full set of response header override options ([the list](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html#RESTObjectGET-requests)), added a test, squashed and rebased.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -56,6 +56,7 @@
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 
+import com.amazonaws.services.s3.model.ResponseHeaderOverrides;
```
andrewgaul: Let's not add a run-time dependency on the SDK just to get these constants.

Dockerfile.build
```diff
@@ -0,0 +1,3 @@
+FROM maven:3-jdk-7-onbuild
+
+CMD ["mvn test"]
```
andrewgaul: Why do we need this `Dockerfile.build` now that S3Proxy provides a `Dockerfile`?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2365,21 +2372,54 @@ private void handleUploadPart(HttpServletRequest request,
...
+            HttpServletResponse response,
+            String headerName,
+            String overrideHeaderName,
+            String value) {
+        String override = request.getParameter(overrideHeaderName);
+        if (override != null) {
+            response.addHeader(headerName, override);
+        } else {
+            response.addHeader(headerName, value);
+        }
```
andrewgaul: Simplify with  `Objects.firstNonNull`.

andrewgaul:
Overall looks good, please address comments so I can merge this.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2388,9 +2428,15 @@ private static void addMetadataToResponse(HttpServletResponse response,
...
-            response.addDateHeader(HttpHeaders.EXPIRES, expires.getTime());
+        String overrideExpires = request.getParameter(
+                ResponseHeaderOverrides.RESPONSE_HEADER_EXPIRES);
+        if (overrideExpires != null) {
+            response.addHeader(HttpHeaders.EXPIRES, overrideExpires);
+        } else {
+            Date expires = contentMetadata.getExpires();
+            if (expires != null) {
+                response.addDateHeader(HttpHeaders.EXPIRES, expires.getTime());
+            }
```
andrewgaul: Similar `Objects.firstNonNull`.

Dockerfile.build
zvikagart: As I explained before this is for a docker based build and development environment.
If you don't think it's useful I'll remove it.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
zvikagart: This one is not straightforward. To avoid unnecessary evaluations you'll have to use the `first.or(supplier)` pattern, plus the fact that you want `expires.getTime()` not just `expires`. I don't see this becoming more readable than it currently is.

andrewgaul: Agree, sorry I misread this.

Dockerfile.build
andrewgaul: Let's remove this for now. This would be useful to get the s3-tests to run locally so I opened #146.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2365,21 +2371,48 @@ private void handleUploadPart(HttpServletRequest request,
...
 
-    private static void addMetadataToResponse(HttpServletResponse response,
+    private static void addResponseHeaderWithOverride(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            String headerName,
+            String overrideHeaderName,
+            String value) {
+        String override = request.getParameter(overrideHeaderName);
+        response.addHeader(headerName, override != null ? override : value);
```
zvikagart: Note that while trying to use `Objects.firstNonNull` I got lots of `NullPointerException` in tests. Probably due to not having a fallback value when calling `Objects.firstNonNull`, probably because some object metadata values were not set.
Instead I used the ternary operator.

andrewgaul:
I made minor stylistic changes, rebased, and pushed to master as fcbc14780b3c45359553848727ac641b499b3e6b.  Thank you for your contribution @zvikagart!

PR #154, opened 2016-09-21
Added logger.error instead of logger.debug in sendSimpleErrorResponse…

kishore25kumar:
This change helps to debug why something is failing from logs

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -2405,7 +2405,7 @@ final class S3ProxyHandler extends AbstractHandler {
             HttpServletRequest request, HttpServletResponse response,
             S3ErrorCode code, String message,
             Map<String, String> elements) throws IOException {
-        logger.debug("{} {}", code, elements);
+        logger.error("{} {}", code, elements);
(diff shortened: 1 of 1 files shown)


andrewgaul:
The debug level is correct; server processes should not log at error when sending an error condition to a client.  Instead error level is for errors within the server, e.g., could not allocate an expected resource, misconfiguration.

kishore25kumar:
At least it should be info. With debug its really hard to find those kind of issues.

andrewgaul:
Arguably this should be trace and not debug; by default server processes should not log individual requests.  We could add some formatting, e.g., `>>>` for requests and `<<<` for responses, to make this easier to match up.

kishore25kumar:
I think without this logs it's very difficult to find out what is happening. As the errors returned by storage service are captured by s3proxy and are sent to the client. That's the reason I was thinking to add it which saves a lot of time. Even if we format the request response it very difficult to see logs and find what is the actual issue. If I enable debug there is a lot of logs printed in the console.

andrewgaul:
I want to follow the following discipline for logging:

https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels

This states that individual response errors fall under the debug category.

PR #163, opened 2016-11-03
Refactor to be used in websphere servlet

shenghu:
3 commits. The purpose is to refactor so that S3ProxyHandlerImpl can be reused in websphere servlet. Please review to see if this can be merged to main stream. BTW, I have successfully use S3ProxyHandlerImpl in my websphere servlet. Although I have more work to do, I think this approach should work and I can rely on S3Proxy to provide core functions. :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,10 @@ FROM anapsix/alpine-java:jre7
 MAINTAINER Andrew Gaul <andrew@gaul.org>
 
 WORKDIR /opt/s3proxy
-COPY target/s3proxy /opt/s3proxy/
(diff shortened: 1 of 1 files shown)


andrewgaul:
@shenghu This is a bit hard to follow -- instead of reparenting all the functionality into `S3ProxyHandlerImpl`, could you rework this to put the Jetty-specific parts in `S3ProxyHandlerJetty`?  This will help preserve git history.  As this is a big change it might help to break this into a few pull requests so we can make progress incrementally.

shenghu:
@andrewgaul sure, I can re-work on it.

PR #165, opened 2016-11-07
Refactor to be used outside of Jetty

shenghu:
Refactor S3ProxyHandler to be agnostic to Jetty. Added a new file, S3ProxyHandlerJetty, to have codes related to Jetty.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/Main.java
+++ b/src/main/java/org/gaul/s3proxy/Main.java
@@ -18,27 +18,14 @@ package org.gaul.s3proxy;
 
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
-import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
-import com.google.common.base.Strings;
-import com.google.common.collect.ImmutableList;
-import com.google.common.io.Files;
-import com.google.inject.Module;
-
-import org.jclouds.Constants;
-import org.jclouds.ContextBuilder;
 import org.jclouds.blobstore.BlobStore;
-import org.jclouds.blobstore.BlobStoreContext;
-import org.jclouds.location.reference.LocationConstants;
-import org.jclouds.logging.slf4j.config.SLF4JLoggingModule;
-import org.jclouds.openstack.swift.v1.blobstore.RegionScopedBlobStoreContext;
 import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
@@ -129,7 +116,7 @@ public final class Main {
         String corsAllowAll = properties.getProperty(
                 S3ProxyConstants.PROPERTY_CORS_ALLOW_ALL);
 
-        BlobStore blobStore = createBlobStore(properties);
+        BlobStore blobStore = S3ProxyHandler.createBlobStore(properties);
 
         Properties altProperties = new Properties();
         for (Map.Entry<Object, Object> entry : properties.entrySet()) {
@@ -144,7 +131,8 @@ public final class Main {
         String eventualConsistency = properties.getProperty(
                 S3ProxyConstants.PROPERTY_EVENTUAL_CONSISTENCY);
         if ("true".equalsIgnoreCase(eventualConsistency)) {
-            BlobStore altBlobStore = createBlobStore(altProperties);
+            BlobStore altBlobStore = S3ProxyHandler
+                    .createBlobStore(altProperties);
             int delay = Integer.parseInt(properties.getProperty(
                     S3ProxyConstants.PROPERTY_EVENTUAL_CONSISTENCY_DELAY,
                     "5"));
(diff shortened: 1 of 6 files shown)


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -211,7 +216,9 @@
             .expireAfterWrite(10, TimeUnit.MINUTES)
             .build();
 
-    S3ProxyHandler(final BlobStore blobStore, final String identity,
+    private BlobStoreLocator blobStoreLocator;
```
andrewgaul: Can you move this back to its original location above?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2352,8 +2349,26 @@ private void handleUploadPart(HttpServletRequest request,
...
-        response.addHeader(headerName, override != null ? override : value);
+
+        // NPE in if value is null
+        override = (override != null) ? override : value;
+
+        if (override != null) {
+            response.addHeader(headerName, override);
+        }
+
+        if (logger.isDebugEnabled()) {
```
andrewgaul: You should not check `isDebugEnabled` since this will optimize out.

src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -206,54 +194,7 @@ public static void main(String[] args) throws Exception {
         }
     }
 
-    private static BlobStore createBlobStore(Properties properties)
```
andrewgaul: Why does this move from `Main` to `S3ProxyHandler` and gain `public visibility?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2437,12 +2452,8 @@ private void sendSimpleErrorResponse(
         }
     }
 
-    public void setBlobStoreLocator(BlobStoreLocator locator) {
-        this.blobStoreLocator = locator;
-    }
-
     @SuppressWarnings("serial")
-    static final class S3Exception extends Exception {
+    public static final class S3Exception extends Exception {
```
andrewgaul: This seems like it should move out of `S3ProxyHandler` now that it is `public`.  I can do this in a subsequent commit or you can include it in this one.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2352,8 +2349,26 @@ private void handleUploadPart(HttpServletRequest request,
     private static void addResponseHeaderWithOverride(
             HttpServletRequest request, HttpServletResponse response,
             String headerName, String overrideHeaderName, String value) {
+
+        if (logger.isDebugEnabled()) {
```
andrewgaul: Is this generally useful or just some temporary logging for development?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2673,14 +2685,18 @@ private static String buildCanonicalHeaders(Request request,
...
-    private static String buildCanonicalQueryString(Request request)
+    private static String buildCanonicalQueryString(HttpServletRequest request)
             throws UnsupportedEncodingException {
         // The parameters are required to be sorted
         List<String> parameters = Collections.list(request.getParameterNames());
         Collections.sort(parameters);
         List<String> queryParameters = new ArrayList<>();
-        String charsetName = Objects.firstNonNull(request.getQueryEncoding(),
+
+        // TODO  Remove this as it is not used?
```
andrewgaul: Agree that the method does not use `charsetName` any more; please remove it in this commit.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
shenghu: I remember I moved it because checkstyle fails. But let me move it back.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
shenghu: This is helpful to debug when using s3proxy as a 3rd party lib.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
shenghu: ok, I remove the check.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
shenghu: done

shenghu:
@andrewgaul can you check the 2nd commit?

shenghu:
@andrewgaul my service will be a micro-service which provide S3 API to inner applications. Our inner applications will use the micro-service to upload documents to object storage. And allow end users to download the documents from the micro-service.

shenghu:
@andrewgaul is it possible to generate a kind of s3proxy-1.5.0-SNAPSHOT-jar-with-dependencies.jar which doesn't contain these ? 

- javax.*
- org.elipse.jetty
- org.slf4j

andrewgaul:
@shenghu It may be possible to slim down the S3Proxy jar a bit, specifically removing the `Main` class.  However most of the dependencies come from jclouds.  Specifically we need to provide Jetty for those using S3Proxy for unit tests.  If you have some ideas on how to narrow the dependencies while preserving this use case please open another pull request!

PR #175, opened 2016-11-29
Allow endpoint of s3proxy to have service path (aka context)

shenghu:
Sometimes service is not the only service under a domain(hostname).
This work adds s3proxy.service-path to allow s3proxy to be deployed
with a context path. The issues resolved by adding service path are
1. properly extract bucket and key, 
2. properly validate signagure
for authentication, 
3. properly validate signagure for presigend url.

Test cases are changed to run when s3proxy.service-path is set.
Right now the setting is commented out so that all tests without
setting it. We need figure out a way to enable it in tests.


shenghu:
@andrewgaul can you have a look at this pull request when you are available?

PR #176, opened 2016-12-08
Make handle x methods overridable

shenghu:
The purpose of making these methods to be overridable is that sub-class want to
leverage S3ProxyHandler to check authentication. And return a presigned url if
a special header is present, instead of doing actual work. This change makes it
possible for sub-class to customize before or after S3ProxyHandler actually
handle the object.

The reason to change pom.xml to use checkstyle 7.3 is to allow setting
'ignoredAnnotations'. Otherwise, it is hard to make handleXYZ methods to be
overridable.

The override method in sub-class may need info in request. Adding
request and response to all handleXYZ methods make them consistent.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,6 +75,13 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
         <version>2.17</version>
+        <dependencies>
+          <dependency>
+            <groupId>com.puppycrawl.tools</groupId>
+            <artifactId>checkstyle</artifactId>
+            <version>7.3</version>
+          </dependency>
+        </dependencies>
         <executions>
           <execution>
             <id>check</id>
--- a/src/main/java/org/gaul/s3proxy/AbstractS3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/AbstractS3ProxyHandler.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright 2014-2016 Andrew Gaul <andrew@gaul.org>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gaul.s3proxy;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.jclouds.blobstore.BlobStore;
+
+
+
+public abstract class AbstractS3ProxyHandler {
+    protected abstract void handleGetContainerAcl(
+            HttpServletResponse response, BlobStore blobStore,
+            String containerName) throws IOException;
+
+    protected abstract void handleSetContainerAcl(
+            HttpServletRequest request, HttpServletResponse response,
+            InputStream is, BlobStore blobStore,
+            String containerName) throws IOException, S3Exception;
+
+    protected abstract void handleGetBlobAcl(
+            HttpServletResponse response, BlobStore blobStore,
+            String containerName,
+            String blobName) throws IOException;
+
+    protected abstract void handleSetBlobAcl(
+            HttpServletRequest request, HttpServletResponse response,
+            InputStream is, BlobStore blobStore,
+            String containerName, String blobName)
+            throws IOException, S3Exception;
+
+    protected abstract void handleContainerList(
+            HttpServletResponse response,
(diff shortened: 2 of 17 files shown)


shenghu:
This PR depends on #175

andrewgaul:
I don't understand the motivation for this change -- why must you make the entire `S3ProxyHandler` class extensible just to override authentication?  Instead shouldn't the `S3Proxy` constructor take some kind of authentication object?

shenghu:
@andrewgaul I think it is good for the entire S3ProxyHandler class to be extensible, which allows consumers to customize it for their specific requirement. And it needn't all the logic to be fed into S3ProxyHandler which might make it hard to maintain in the future. I mean S3ProxyHandler is good and should keep to be concise as much as possible for core functions. A 3rd party can use its extensibility to test their code. If they think their codes work well and are common enough, they can consider to contribute back here. 

For me, the initial motivate is to ask S3Proxy to generate pre-signed URL for object storage so that I can redirect users to object storage w/o going through S3Proxy (going through S3Proxy is not able to scale horizontally in my project). Although right now just download request need this capability, it could be used by upload request. So what I do is , e.g. 

`  @Override
  protected void handleGetBlob(HttpServletRequest request, HttpServletResponse response, BlobStore blobStore, String containerName,
      String blobName) throws IOException, S3Exception
  {
    if (this.isPresignedURLGenerationRequest(request))
    {
      this.handlePresignURLGenerationRequest(request, response, containerName, blobName);
      return;
    }

    super.handleGetBlob(request, response, blobStore, containerName, blobName);
  }

  @Override
  protected void handleGetBlobAcl(HttpServletRequest request, HttpServletResponse response, BlobStore blobStore, String containerName,
      String blobName) throws IOException
  {
    if (this.isPresignedURLGenerationRequest(request))
    {
      this.handlePresignURLGenerationRequest(request, response);
      return;
    }

    super.handleGetBlobAcl(request, response, blobStore, containerName, blobName);
  }`

And in handlePresignURLGenerationRequest(), it does this
`protected void handlePresignURLGenerationRequest(HttpServletRequest request, HttpServletResponse response, String containerName,
    String key) throws IOException
{
  Map<Constants.OverrideResponseHeader, String> overridedHeaders = HttpUtils.getOverrideResponseHeaders(request);

  BasicAWSCredentials awsCreds = new BasicAWSCredentials("username", "password");

    String presignedUrl = BlobUtils.generateStoragePresignedUrl(HttpMethod.valueOf(request.getMethod()), awsCreds, containerName, key,
        expiration,
        signerType,
        pathStyleAccess, 
        overridedHeaders,
        endpoint);

    // Set status code to be 204
    response.setStatus(HttpServletResponse.SC_NO_CONTENT);

    // Write content as pre-signed url.
    response.setHeader(Constants.HEADER_STORAGE_PRESIGNED_URL, presignedUrl);
  }
  catch (Exception e)
  {
    // log Exception
  }
}`

When pre-signed URL is generated, it need some configurations like expiration, path style. I'm not sure how to add them into S3ProxyHandler. This way looks the best for me to leverage S3ProxyHandler and do my customization. 

Please let me know your thought.

andrewgaul:
I see what you are doing here and think you could improve on it in a more generic way.  Instead of passing back a custom header, what if you returned at HTTP 307 redirect with the presigned URL?  This would allow all conformant HTTP clients to work.  HDFS uses a similar approach to redirect requests from master to slaves.

Generally I like your feature and wonder if we could make it part of S3Proxy itself?  I don't think we can use the `BlobStore` middleware approach directly since you write custom response code and headers.  However we could have a custom exception type which tells S3Proxy to emit a redirect with a given URL that the middleware could provide.  Making S3Proxy more extensible is a goal for me and I wonder if something like #181 would help?

We can also go the route of making more of the S3Proxy interface public but I prefer to explore this other approach first.  What do you think?

shenghu:
@andrewgaul the reason I use a customer header instead of 307 is that the route is client->myService->S3proxy. I don't want myService to follow redirect, but just to get the url and give it to client w/ 302/307.

andrewgaul:
Can you configure your service to not follow redirects then vend the response back to the client?  Most HTTP clients allow this.

shenghu:
@andrewgaul I can try to merge the code about generating pre-signed url to s3proxy and return 307. I need some time to test w/ my codes. 

Another thing is I met a problem recently which makes me think it is still better to make s3ProxyHandler to be extendable. The issue is that blobMeta method return object size using Content-Length header. When the value is larger than Integer.MAX_VALUE, whether Content-Length can successfully set the value depends on web container. Since we are using Websphere 8.5.x which use integer internally, it doesn't set Content-Length if it is larger than Integer.MAX_VALUE. Fortunately, I can override handleBlogMetadata method and set a customer header to work around the bug of container. In general, I hope you agree to make S3ProxyHandler to be extendable so that it gives consumers more options.

andrewgaul:
@shenghu Do we have a path forward on this pull request?  Writing a middlware for 307 redirect seems clean and useful to others.

andrewgaul:
I opened #227 to track the 307 redirect middleware.

shenghu:
@andrewgaul sorry, I was busy on a different project. 307 redirect can not be handled well in my client which use jclouds as client. It might be better that s3proxy can be configured to do 307 or return 200 + special header.

andrewgaul:
Do you mean that jclouds does not handle 307 redirects?  If so, please [file an issue](https://issues.apache.org/jira/projects/JCLOUDS) so jclouds can track this.

PR #178, opened 2016-12-17
S3proxy presigned url fail with override parameters

shenghu:
This change makes S3Proxy able to validate presigned url which has override parameters, i.e. "content-disposition", "response-content-encoding".

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -110,7 +110,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /** HTTP server-independent handler for S3 requests. */
-public class S3ProxyHandler {
+public class S3ProxyHandler extends AbstractS3ProxyHandler {
     private static final Logger logger = LoggerFactory.getLogger(
             S3ProxyHandler.class);
     private static final String AWS_XMLNS =
@@ -672,7 +672,8 @@ public class S3ProxyHandler {
         throw new S3Exception(S3ErrorCode.NOT_IMPLEMENTED);
     }
 
-    private void handleGetContainerAcl(HttpServletResponse response,
+    @Override
+    protected void handleGetContainerAcl(HttpServletResponse response,
             BlobStore blobStore, String containerName) throws IOException {
         ContainerAccess access = blobStore.getContainerAccess(containerName);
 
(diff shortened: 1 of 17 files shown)


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -140,7 +140,14 @@
...
             "partNumber", "policy", "requestPayment", "torrent", "uploadId",
-            "uploads", "versionId", "versioning", "versions", "website"
+            "uploads", "versionId", "versioning", "versions", "website",
+            // Added from com.amazonaws.services.s3.internal.RestUtils
+            "tagging", "cors", "restore", "replication", "accelerate",
+            "inventory", "analytics", "metrics",
+            // Override-response-header parameters aslo need be signed
+            "response-content-type", "response-content-language",
+            "response-expires", "response-cache-control",
+            "response-content-disposition", "response-content-encoding"
```
andrewgaul: Can you only include the `response-*` parameters since S3Proxy does not support ones like `cors` and similar.  Also please keep the alphabetization.

src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
```diff
@@ -135,8 +136,9 @@ public void tearDown() throws Exception {
         }
     }
 
+
     @Test
-    public void testAwsV2Signature() throws Exception {
+    public void testAwsV2SignatureWithOverrideParameters() throws Exception {
```
andrewgaul: Can you break this into its own test?  I want to retain the existing `testAwsV2Signature`.

src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
```diff
@@ -145,15 +147,34 @@ public void testAwsV2Signature() throws Exception {
...
+        assertThat(object.getObjectMetadata().getContentType()).isEqualTo(
+                expectedContentType);
         try (InputStream actual = object.getObjectContent();
-                InputStream expected = BYTE_SOURCE.openStream()) {
+             InputStream expected = BYTE_SOURCE.openStream()) {
             assertThat(actual).hasContentEqualTo(expected);
         }
     }
 
+
```
andrewgaul: Remove newline.

src/test/java/org/gaul/s3proxy/S3AwsSdkTest.java
```diff
@@ -253,6 +274,39 @@ public void testAwsV2UrlSigning() throws Exception {
         }
     }
 
+    @Test
+    public void testAwsV2UrlSigningWithOverrideParameters() throws Exception {
```
andrewgaul: Does this actually give some coverage that `testAwsV2SignatureWithOverrideParameters` does not?  I usually rely on [s3-tests](https://github.com/ceph/s3-tests) for this kind of exhaustive testing.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
shenghu: ok

shenghu:
This is re-worked by #187

Issue #160, opened 2016-10-18
Support V4 Signed URLs


andrewgaul:
Current error:

```
E 10-18 13:06:32.338 S3Proxy-24 o.gaul.s3proxy.S3ProxyHandler:480 |::] Unknown parameters X-Amz-Expires with URI /temporary-bucket/68ae4a90-f079-02d1-6b06-fa3cd4e3aa26/c697097d-aad2-4d8c-293c-a4159f0a9286
```

Originally reported by @historus in #138.

andrewgaul:
Currently ignoring `S3AwsSdkTest.testAwsV4UrlSigning` test.

PR #186, opened 2017-02-20
Add support for ignoring unknown parameters

### LINKED ISSUE #138: Provide ability to also ignore unknown parameters
There is already an option to ignore unknown headers.
It would be helpful to also ignore unknown parameters.

smigfu:
Presigned URLs may contain additional parameters in the URL which will
be used “as is” by the client uploading/downloading content from S3.

CLOSES #138

### DIFF
--- a/src/main/java/org/gaul/s3proxy/Main.java
+++ b/src/main/java/org/gaul/s3proxy/Main.java
@@ -134,6 +134,8 @@ public final class Main {
                 S3ProxyConstants.PROPERTY_V4_MAX_NON_CHUNKED_REQUEST_SIZE);
         String ignoreUnknownHeaders = properties.getProperty(
                 S3ProxyConstants.PROPERTY_IGNORE_UNKNOWN_HEADERS);
+        String ignoreUnknownParameters = properties.getProperty(
+                S3ProxyConstants.PROPERTY_IGNORE_UNKNOWN_PARAMETERS);
         String corsAllowAll = properties.getProperty(
                 S3ProxyConstants.PROPERTY_CORS_ALLOW_ALL);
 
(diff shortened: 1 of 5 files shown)


andrewgaul:
Instead of hacking around this, why not address #160 instead?

smigfu:
I don't know if adding the v4 signing support would help? The PUT/POST/DELETE call which is done on the presigned url returned to the client would still have the parameters not listed in the SUPPORTED_PARAMETERS list in S3ProxyHandler.

Let me clarify the workflow I am trying to support:

1. client calls my service asking for a presigned url to upload content
2. my service generates the presigned url and adds some custom parameters (customer_id, etc)
3. client invokes POST on the presigned url (which includes the paramters now)

=> s3proxy throws the unknown parameter exception on the POST for the content

andrewgaul:
Sorry I am missing something here -- which parameter does S3Proxy fail with?  What does Amazon do with custom parameters like `customer_id`?  I thought it would fail those requests.  Are we talking about user metadata?  I am trying to determine if S3Proxy needs to improve its presigned URL support or if there is another issue.

smigfu:
I can request a presigned URL via the Java API. This requests is handled within the Java library, no call to AWS is done. I can add arbitrary parameters to this URL, i.e.:

```
https://<s3-endpoint>/my-bucket/my-object?<aws-specific-parameters>&customer_id=abc

(customer_id=abc is my custom parameter I want to see in this URL)
```

This URL is given to my customer which in turn will upload his content to this URL via POST. AWS just ignores the customer_id parameter in that case. But s3proxy will fail because customer_id is not a supported parameter (contained in the SUPPORTED_PARAM_LIST).

andrewgaul:
I had not considered this -- you add `customer_id` which Amazon ignores but gets included in some kind of logging?  S3Proxy strives for compatibility with Amazon so it should also ignore unknown parameters by default.  However this requires flipping the logic and including a list of explicitly unsupported parameters, e.g., `X-Amz-Algorithm`.  This will require some research so I opened #189 and will merge this pull request in the mean time.

andrewgaul:
Thank you for your contribution @smigfu!

smigfu:
Great, thank you! 👍

Issue #121, opened 2016-02-06
Java 8 dependency


gaul:
S3Proxy requires Java 7 presently.  Requiring Java 8 would allow upgrading some dependencies:

- Checkstyle 7.0 http://checkstyle.sourceforge.net/releasenotes.html#Release_7.0
- error-prone 2.0.6 https://groups.google.com/forum/#!topic/error-prone-announce/ycJBeaNK3Uc
- Jetty 9.3 https://webtide.com/jetty-9-3-features/

These do not seem compelling and older distributions like Ubuntu 14.04 LTS do not provide Java 8.  Keeping this bug open for tracking.

rbjorklin:
A Java 8 PPA can be found here though: http://www.webupd8.org/2012/09/install-oracle-java-8-in-ubuntu-via-ppa.html

andrewgaul:
Moving to JDK 8 would enable using Nashorn for JavaScript middlewares discussed in #181.

@shenghu @SpandanThakur You both spoke about the JDK 8 dependency in #190.  Any opposition to moving now?

SpandanThakur:
@andrewgaul No opposition. I think moving to JDK 8 would be a good move.

gaul:
jclouds 2.2.0 will require Java 8 so S3Proxy will move forward with it.

gaul:
S3Proxy 1.8.0 requires Java 8 so we can upgrade these dependencies now.

PR #190, opened 2017-03-01
changes for java 1.8

SpandanThakur:
Changes done to move to java 1.8.

This project uses a moderniser-maven-plugin which showed the below issues. I resolved them in this pull request.

```
[ERROR] /Users/sthakur/s3proxy/src/main/java/org/gaul/s3proxy/S3Proxy.java:106: Prefer java.util.Optional
[ERROR] /Users/sthakur/s3proxy/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java:1583: Prefer java.util.Base64.Decoder or java.util.Base64.Encoder
[ERROR] /Users/sthakur/s3proxy/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java:1759: Prefer java.util.Base64.Decoder or java.util.Base64.Encoder
[ERROR] /Users/sthakur/s3proxy/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java:2233: Prefer java.util.Base64.Decoder or java.util.Base64.Encoder
[ERROR] /Users/sthakur/s3proxy/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java:2519: Prefer java.util.Base64.Decoder or java.util.Base64.Encoder
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -272,7 +272,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <java.version>1.7</java.version>
+    <java.version>1.8</java.version>
     <jclouds.version>2.0.1</jclouds.version>
     <slf4j.version>1.7.23</slf4j.version>
     <surefire.version>2.19.1</surefire.version>
--- a/src/main/java/org/gaul/s3proxy/S3Proxy.java
+++ b/src/main/java/org/gaul/s3proxy/S3Proxy.java
@@ -21,8 +21,8 @@ import static java.util.Objects.requireNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import java.net.URI;
+import java.util.Optional;
 
-import com.google.common.base.Optional;
 import com.google.common.base.Strings;
 
 import org.eclipse.jetty.server.HttpConnectionFactory;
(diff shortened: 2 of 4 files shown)


andrewgaul:
What benefit does requiring Java 8 give users?  #121 discusses some dependencies.

shenghu:
I hope s3proxy not to require java8 now, which can give s3proxy more changes to be used. I think jclouds also just require java7. S3Proxy needn't move unless jclouds moves.

SpandanThakur:
Hi,
I done this change because we are planning to contribute to jClouds as well for asynchronous behaviour. I agree though that we can keep this on hold till we contribute in jClouds to make it on java8 as well.

I posted regarding the change that we are planning in jClouds on the jclouds forum but I am pasting below as well. Any suggestions will be welcome.

"We are thinking and planning to add proper multiplexing Async behaviour in jClouds. We plan to use the apache async httpClient library to achieve the same.  We are still thinking about how to implement this in jClouds and s3 proxy but below are are initial thoughts. 
1. Move to Java 1.8 so we can start using CompletableFutures. (Let us know if you feel there are any issue in upgrading JAVA version) 
2. Add a new annotation like @async which can be used in the client interfaces like AzureBlobClient.java, S3Client.java. 
3. Add new methods for async in BlobStore and implement then in the specific stores. The methods will return completableFutures. 
4. In the core http package add the apache async httpClient based class. Which will execute async http requests and return completable futures. (We are not sure if we can just add a implementation of BaseHttpCommandExecutorService.java for the same) 
5. Make some changes in the DelegatesToInvocationFunction.java to see if async annotation is present and then using the Async client http executor. 
6. The S3 proxy use jetty continuations to suspend the thread and then continue after the async method is done (using thenAccept in the Completable Future).  
Please do let me know your thoughts on this. Do you feel like this feature would be something useful for jClouds? Do you think we are on the right track on the way we plan to implement it? Any suggestions from your side?"

andrewgaul:
Closing this for now.  When jclouds requires Java 8 S3Proxy will update its requirements as well.

PR #194, opened 2017-03-04
Support service path

shenghu:
Sometimes service is not the only service under a domain(hostname). This work adds s3proxy.service-path to allow s3proxy to be deployed with a service path. The issues resolved by adding service path are 1) properly extract bucket and key, 2) properly validate signagure for authentication, 3) properly validate signagure for presigned url. Test cases are changed to run when s3proxy.service-path is set.


shenghu:
@andrewgaul How about this work?

andrewgaul:
Thank you for your contribution @shenghu!

PR #207, opened 2017-03-30
Update run-docker-container.sh

mcanevet:


andrewgaul:
Merged #208 instead which sets `JCLOUDS_REGION` in the `Dockerfile` and has an actual commit message.

PR #210, opened 2017-04-25
Update docker script to allow for configuration of virtualhost

hydrogen18:


andrewgaul:
Thank you for your contribution @hydrogen18!

src/main/resources/run-docker-container.sh
```diff
@@ -3,6 +3,7 @@
 exec java \
     -DLOG_LEVEL=${LOG_LEVEL} \
     -Ds3proxy.endpoint=http://0.0.0.0:80 \
+    -Ds3proxy.virtual-host=${S3PROXY_VIRTUALHOST} \
```
jim-slattery-rs: our builds using the s3proxy dockerhub image started breaking when this change was introduced..

maybe `virtual-host` used to be null? is there any way for it to still be null if it's included in the command line like this?

hydrogen18: Perhaps we just need to have the configuration methods consider `null` and the empty string to be the same thing.

andrewgaul:
@jim-slattery-rs 6b30e6612779e54c1741e522a09d6bd6b2dd31b4 should fix this and Travis updated the Docker image.  Sorry about the breakage and thanks for your feedback!

src/main/resources/run-docker-container.sh
andrewgaul: @jim-slattery-rs 6b30e6612779e54c1741e522a09d6bd6b2dd31b4 should fix this and Travis updated the Docker image.  Sorry about the breakage and thanks for your feedback!

PR #215, opened 2017-05-02
Fixed failing test in case of azure

kishore25kumar:
For blob stores for which there is no blob access control, testHttpClient test is failing. For these blob stores by default getBlobAccess returns private. If a bucket is public then we should allow public access of the blob. This pull request fixes this issue

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -658,9 +658,18 @@ public class S3ProxyHandler {
             } else {
                 String containerName = path[1];
                 String blobName = path[2];
-                BlobAccess access = blobStore.getBlobAccess(containerName,
-                        blobName);
-                if (access != BlobAccess.PUBLIC_READ) {
+                boolean publicAccess = false;
+                String blobStoreType = getBlobStoreType(blobStore);
+                if (Quirks.NO_BLOB_ACCESS_CONTROL.contains(blobStoreType)) {
+                    ContainerAccess access = blobStore.getContainerAccess(
+                            containerName);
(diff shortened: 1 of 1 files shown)


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -658,9 +658,18 @@ private void doHandleAnonymous(HttpServletRequest request,
...
+                if (Quirks.NO_BLOB_ACCESS_CONTROL.contains(blobStoreType)) {
+                    ContainerAccess access = blobStore.getContainerAccess(
+                            containerName);
+                    publicAccess = access == ContainerAccess.PUBLIC_READ;
+                } else {
+                    BlobAccess access = blobStore.getBlobAccess(containerName,
+                            blobName);
+                    publicAccess = access == BlobAccess.PUBLIC_READ;
+                }
+                if (!publicAccess) {
```
andrewgaul: Replicate this logic for HEAD object below.  Factor out the check into a function named `checkPublicAccess`.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -658,9 +658,18 @@ private void doHandleAnonymous(HttpServletRequest request,
             } else {
                 String containerName = path[1];
                 String blobName = path[2];
-                BlobAccess access = blobStore.getBlobAccess(containerName,
-                        blobName);
-                if (access != BlobAccess.PUBLIC_READ) {
+                boolean publicAccess = false;
```
andrewgaul: Remove unneeded default value.

andrewgaul:
Thank you for your contribution @kishore25kumar!

PR #218, opened 2017-05-05
Add read-only BlobStore middleware

gaul:
Implement as a proof-of-concept JavaScript middleware.  Requires JDK
8.

### DIFF
--- a/src/test/resources/readonly_blobstore.js
+++ b/src/test/resources/readonly_blobstore.js
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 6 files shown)


src/test/resources/readonly_blobstore.js
```diff
@@ -0,0 +1,64 @@
...
+        },
+        clearContainer: function(container) {
+            throw new java.lang.UnsupportedOperationException();
+        },
+        setContainerAccess: function(container, access) {
+            throw new java.lang.UnsupportedOperationException();
+        },
+
+        // object operations
+        putBlob: function(containerName, blob) {
```
andrewgaul: How does this work for the three-argument `putBlob` method?

gaul:
I have soured on JavaScript middlewares for now but would like to rewrite the read-only middleware and merge it.

gaul:
Merged Java read-only middleware in 35d7c4847570046ab1aeafbbb3e3dd8228c5bd21.

Issue #219, opened 2017-05-05
Asynchronously stream data


andrewgaul:
S3Proxy could reduce its thread usage by asynchronously streaming data instead of synchronously copying it.  `BlobStore.getBlob` provides an `InputStream` which allows this but `BlobStore.putBlob` does not expose an `OutputStream` to write to, tracked by [JCLOUDS-769](https://issues.apache.org/jira/browse/JCLOUDS-769).  With both of these, we could use the standard async Servlet interface following the Jetty example here:

https://webtide.com/servlet-3-1-async-io-and-jetty/

@SpandanThakur related to your discussion on jclouds-dev.

gaul:
Revisiting this nine years later: the goal still stands, but the prescribed mechanism no longer does.

* JCLOUDS-769 will never be fixed — jclouds is in the Apache Attic, and S3Proxy is sunsetting the jclouds backends after 3.3.0 anyway. The `BlobStore` interface is now a fork under this project's control, so the missing streaming API stopped being the blocker. That no longer matters, though, because:
* Rewriting `S3ProxyHandler` for Servlet async I/O would turn roughly 4,000 lines of synchronous logic — SigV4 `aws-chunked` parsing, checksum trailers, multipart upload, conditional requests — into callback state machines. That is a large regression surface for a benefit the platform now provides directly.
* Virtual threads (JEP 444, Java 21) deliver what this issue was after — threads stop being a scarce resource during blocking I/O — with no changes to request-handling code, and JEP 491 (Java 24) removed the `synchronized` pinning caveat that would have mattered for the older client libraries in the dependency tree. The Jetty 12.1 that S3Proxy bundles already ships `VirtualThreadPool` and `QueuedThreadPool.setVirtualThreadsExecutor`, so the eventual change is small: swap the `QueuedThreadPool` in `S3Proxy` for a `VirtualThreadPool`, then validate with a load test (several thousand slow concurrent GETs, comparing thread count, RSS, and latency).

S3Proxy targets Java 17 as its floor and I am not ready to take a Java 21 dependency yet. An opt-in pool would keep `--release 17` compiling and only require 21+ at runtime when enabled, but I would rather not carry a runtime-conditional code path for this. Parking the issue until the Java floor moves to 21 or later; at that point virtual threads, not Servlet async I/O, are the way to implement it.

PR #220, opened 2017-05-05
Async response proof of concept

gaul:
This streams data during Get Object requests.  References #219.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -23,7 +23,6 @@ import java.io.ByteArrayOutputStream;
 import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
 import java.io.PushbackInputStream;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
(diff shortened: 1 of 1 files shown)


andrewgaul:
@SpandanThakur Please review.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2871,4 +2873,38 @@ private static boolean validateIpAddress(String string) {
...
+                ServletOutputStream sos) {
+            this.async = async;
+            this.is = is;
+            this.sos = sos;
+        }
+
+        public void onWritePossible() throws IOException {
+            byte[] buffer = new byte[4096];
+            while (sos.isReady()) {
+                int len = is.read(buffer);
```
kishore25kumar: @andrewgaul this won't solve the problem we are addressing. This will just write the response to the client in an asynchronous fashion. Even in this case is.read(buffer) is a blocking call. What we wanted to solve is when we make a request to put blob or get blob we will return a listenable future which will be completed only when there is a response from the service provider.

kishore25kumar: @andrewgaul any update on this

andrewgaul: I would like to see the benchmarks that show that this approach is significantly worse in real-world tests.  You will need to do all the legwork on this.  Please update jclouds-dev with your results.

gaul:
Closing due to inactivity.

PR #225, opened 2017-05-24
Validate bucket name before processing any request

ChaithanyaGK:


andrewgaul:
Can you add a test to [s3-tests](https://github.com/ceph/s3-tests) which exercises this?

ChaithanyaGK:
@andrewgaul  I've raised a [PR](https://github.com/ceph/s3-tests/pull/164) with a test case to s3-tests and will try to add few more soon and made minor changes to throw proper S3 error codes, to be in line with AWS S3

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -547,6 +559,17 @@ public final void doHandle(HttpServletRequest baseRequest,
             }
         }
 
+        // Validate container name
+        if (!uri.equals("/") && !isValidContainer(path[1])) {
+            if (method.equals("PUT") &&
+                    (path.length <= 2 || path[2].isEmpty()) &&
+                    !("".equals(request.getParameter("acl"))))  {
```
andrewgaul: `getParameter` can return `null`.  Instead call `Strings.isNullOrEmpty`?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -547,6 +559,17 @@ public final void doHandle(HttpServletRequest baseRequest,
...
+                throw new S3Exception(S3ErrorCode.INVALID_BUCKET_NAME);
+            } else {
+                throw new S3Exception(S3ErrorCode.NO_SUCH_BUCKET);
```
andrewgaul: The underlying `BlobStore` operations should throw `ContainerNotFoundException`.  Unfortunately the jclouds does not propagate this.  Could you look at [JCLOUDS-193](https://issues.apache.org/jira/browse/JCLOUDS-193)?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
ChaithanyaGK: It can return null( if "acl" is not present), But, we are here checking for an empty "acl" value. Please [refer](https://github.com/andrewgaul/s3proxy/blob/742820a93d4d87e3ff249ae85e92f3eaf405db6e/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java#L622)

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
ChaithanyaGK: Yes, with this commit at least we will throw a proper exception in the case of bad bucket name irrespective of `BlobStore` operations

ChaithanyaGK:
@andrewgaul  Can this be merged?

andrewgaul:
Thank you for your contribution @ChaithanyaGK!

Issue #146, opened 2016-07-14
Provide Docker build environment


andrewgaul:
This would run the unit tests but more importantly run the s3-tests conformance suite.  This would allow developers to easily run the same tests as Travis.

PR #228, opened 2017-06-01
Update Dockerfile to use multistage building.

nepeat:
This allows developers and users to build their own version of s3proxy fully in Docker without the need of external builders for the s3proxy binary.

This pull depends on features that have been recently released in the [17.05](https://github.com/moby/moby/blob/17.05.x/CHANGELOG.md#builder) version of Docker which may not be preferable for some older build environments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,22 @@
-FROM anapsix/alpine-java:jre7
+# Multistage - Builder
+FROM maven:3.5.0-jdk-7-alpine as s3proxy-builder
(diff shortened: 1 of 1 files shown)


andrewgaul:
Multistage building seems like a good way to address #146 and simplify the Travis configuration.  I am on a long holiday but I will return to this in a few weeks.  Thanks for raising this pull request!

gaul:
@nepeat Thank you for your contribution and educating me about multi-stage builds!  Sorry I took so long to merge this; I got stuck trying to include s3-tests and forgot about it for some time.

PR #230, opened 2017-06-06
Emit proper S3 url/location in multipart upload response

ChaithanyaGK:


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2039,9 +2039,11 @@ private void handleCompleteMultipartUpload(HttpServletResponse response,
             xml.writeStartElement("CompleteMultipartUploadResult");
             xml.writeDefaultNamespace(AWS_XMLNS);
 
-            // TODO: bogus value
-            writeSimpleElement(xml, "Location",
-                    "http://Example-Bucket.s3.amazonaws.com/" + blobName);
+            String s3Amazon = "s3.amazonaws.com/";
```
andrewgaul: Should this be `builder.endpoint` or `builder.secureEndpoint` from `S3Proxy`?

ChaithanyaGK: Yes, builder.endpoint or builder.secureEndpoint is more relevant. Let me put in some more thought and will raise a clean pull request.

Closing this PR.

PR #233, opened 2017-06-20
S3Proxy Code refactoring POC

ChaithanyaGK:
Signed-off-by: Chaithanya Ganta <ganta@adobe.com>


ChaithanyaGK:
This PR is just to give an overview of the design on refactoring the S3ProxyHandler.

src/main/java/org/gaul/s3proxy/S3Exception.java
```diff
@@ -27,33 +27,33 @@
     private final S3ErrorCode error;
     private final Map<String, String> elements;
 
-    S3Exception(S3ErrorCode error) {
+    public S3Exception(S3ErrorCode error) {
```
andrewgaul: Why make these public visibility?  I don't think the requests need their own subpackage so you can move them into the main package and reduce visibility here.

src/main/java/org/gaul/s3proxy/requesthandlers/S3BucketListHandler.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamWriter;
+
+import org.gaul.s3proxy.S3Exception;
+import org.gaul.s3proxy.XMLUtils;
+import org.jclouds.blobstore.BlobStore;
+import org.jclouds.blobstore.domain.PageSet;
+import org.jclouds.blobstore.domain.StorageMetadata;
+
+public final class S3BucketListHandler extends AbstractS3BucketHandler {
```
andrewgaul: I do not think using objects actually adds anything here.  Instead we can implement these as static utility methods.  I envision them in larger buckets to start, e.g., `BucketRequest`, `ObjectRequest`, `MultipartRequest`, mirroring the jclouds features.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -18,6 +18,12 @@
 
 import static java.util.Objects.requireNonNull;
 
+// CHECKSTYLE:OFF
```
andrewgaul: I understand why these static imports are temporarily needed but please add a TODO to remove them when all methods are reparented.

src/main/java/org/gaul/s3proxy/requesthandlers/S3BucketListHandler.java
ChaithanyaGK: 1. Always using objects are preferred over static utility methods in an object-oriented model. In fact, Jclouds also uses objects for these operations.
2. If we categorize them as larger entities like bucket or object, it will again increase the complexity of the code if we keep on adding new bucket or object functionalities in future.

src/main/java/org/gaul/s3proxy/S3Exception.java
ChaithanyaGK: Agreed, will move them into the main package

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
ChaithanyaGK: Sure, will add a TODO

src/main/java/org/gaul/s3proxy/requesthandlers/S3BucketListHandler.java
andrewgaul: I do not share you views on code style.  Please address my comment as previously written.

ChaithanyaGK:
@andrewgaul I've made changes as requested. Could you please have a look.

ChaithanyaGK:
@andrewgaul Could you please review the design.

gaul:
Closing this since it has become stale.  Sorry I was not more responsive but I would be more comfortable doing large-scale refactoring myself.

PR #238, opened 2017-08-02
Emit InvalidArgument on CopyPart invalid range

andrewgaul:
References kahing/goofys#212.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -2174,6 +2174,7 @@ public class S3ProxyHandler {
 
         GetOptions options = new GetOptions();
         String range = request.getHeader("x-amz-copy-source-range");
+        long expectedSize = -1;
         if (range != null && range.startsWith("bytes=") &&
                 // ignore multiple ranges
                 range.indexOf(',') == -1) {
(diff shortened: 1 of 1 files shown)


andrewgaul:
@kahing Please review.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2221,6 +2224,10 @@ private void handleCopyPart(HttpServletRequest request,
         }
 
         BlobMetadata blobMetadata = blob.getMetadata();
+        // HTTP GET allow overlong ranges but S3 CopyPart does not
+        if (expectedSize != -1 && blobMetadata.getSize() < expectedSize) {
```
kahing: what happens if I request 5-6 for a blob that has 4 bytes?

andrewgaul: Normal GET handling will reject the invalid beginning.  This code only needs to add the additional CopyPart checks.

kahing: what about 4-6 for a blob that's 5 bytes?

andrewgaul: The size check will catch that.

kahing: how so? expectedSize = 6-4+1 = 3, blobMetadata.getSize() will be 5.

andrewgaul: S3Proxy passes the range to `getBlob` so that getSize will return 1 here.

PR #239, opened 2017-08-03
Do not escape / when URL encoding

andrewgaul:
This allows clients to list pseudo-directories with slashes in them.
References kahing/goofys#213.


andrewgaul:
@kahing Please review.

PR #257, opened 2018-01-24
Update README.md

dyusupov:


gaul:
The references contain S3 implementations which translate from S3 to other protocols, e.g., Azure, Swift, POSIX.  edgefs translates from POSIX to S3, similar to s3fs, and the reverse of what S3Proxy does.  Do I understand this correctly?  If so please close this pull request since I do not want to enumerate the many FUSE S3 implementations.

dyusupov:
Thanks for quick reply. EdgeFS translates S3 access to any of the protocols
NexentaEdge provides, which includes: OpenStack/SWIFT, NFS and even iSCSI
or NBD.

gaul:
Do you mean that EdgeFS translates from POSIX to S3 and NexentaEdge translates from S3 to other protocols?  If the NexentaEdge protocol translator is proprietary software I do not want to add an advertising reference.  Neither of these projects have any kind of license file and it appears that you provide some kind of license key.

PR #265, opened 2018-02-23
Shade and relocate Jetty dependencies

### LINKED ISSUE #263: Jetty dependency issues
We have been trying to use S3Proxy to test code in a project which also uses Hive. Hive requires Jetty 7.x while S3Proxy requires Jetty 9.x. We have certain use cases where the tests are impossible to run as obviously only one version of Jetty can be on the classpath and we get various exceptions due to missing Jetty classes, incompatible method definitions etc. Would it be possible to shade and relocate the Jetty classes inside S3Proxy so the usage of it is effectively hidden from downstream users? I'm not a huge fan of shading but in situations like this it can be really helpful for allowing integration into projects which depend on different versions of a required library.

massdosage:
Fixes #263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+## [Unreleased]
+### Changed
+* `org.eclipse.jetty.*` classes are now shaded and relocated into main s3proxy jar (fixes [#263](https://github.com/gaul/s3proxy/issues/263))
+
+## [1.5.5] - 2018-01-25
+* This and previous versions didn't keep a [changelog](http://keepachangelog.com), refer to [releases](https://github.com/gaul/s3proxy/releases) for details instead.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,0 @@
-## [Unreleased]
-### Changed
-* `org.eclipse.jetty.*` classes are now shaded and relocated into main s3proxy jar (fixes [#263](https://github.com/gaul/s3proxy/issues/263))
-
-## [1.5.5] - 2018-01-25
-* This and previous versions didn't keep a [changelog](http://keepachangelog.com), refer to [releases](https://github.com/gaul/s3proxy/releases) for details instead.
--- a/src/main/assembly/jar-with-dependencies.xml
+++ b/src/main/assembly/jar-with-dependencies.xml
@@ -13,7 +13,10 @@
   </containerDescriptorHandlers>
   <dependencySets>
     <dependencySet>
-      <outputDirectory>/</outputDirectory>
+      <excludes>
+        <exclude>org.eclipse.jetty:*</exclude>
+      </excludes>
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+## [Unreleased]
```
massdosage: I noticed S3Proxy doesn't have a changelog so I thought I would add one so one can capture documenting changes as part of pull request.

src/main/assembly/jar-with-dependencies.xml
```diff
@@ -22,7 +25,7 @@
   <fileSets>
     <fileSet>
       <directory>${project.basedir}/src/main/config</directory>
-      <outputDirectory>/</outputDirectory>
+      <outputDirectory/>
```
massdosage: This doesn't have anything to do with my change but when I was running the "package" phase in Maven it complained that  "/" wasn't cross-platform. Specifying an empty string for "outputDirectory" (which this effectively does) seems to achieve the intended result of locating files at the root of the package, without the warning.

massdosage:
I'm not sure why the build is failing in Travis. The Maven side of things (which I tested locally) is fine but something is going wrong in the Docker phase. I haven't tested this as we're not using S3Proxy in this way and I wasn't sure what I needed to do. I tried to make sure that the "-all" jar file was "correct" but it's possible I missed something there.

CHANGELOG.md
gaul: Please remove.  The release page duplicates this:

https://github.com/gaul/s3proxy/releases

src/main/assembly/jar-with-dependencies.xml
gaul: Please remove.  Instead open separate pull requests for distinct changes.

src/main/config/logback.xml
```diff
@@ -9,6 +9,7 @@
   </appender>
 
   <logger name="org.eclipse.jetty" level="${JETTY_LOG_LEVEL:-info}" />
+  <logger name="org.gaul.shaded.org.eclipse.jetty" level="${JETTY_LOG_LEVEL:-info}" />
```
gaul: Why have both of these?

massdosage: I don't know enough about how the "-all" jar is run as we don't use it that way so I wasn't able to test this. I assume only the line I added is needed since the classes are now all relocated. This also assumes Jetty have named their loggers after class names, if not then you'd also need the first line.

massdosage:
Yes, it should be transparent to users. I tested this in the downstream build of ours which had the issue with incompatible versions of Jetty and didn't have to change anything and the new version solved my problem. If there are issues with this going forward I'm happy to try help solve them.

gaul:
Thank you for your contribution @massdosage!

Issue #237, opened 2017-07-25
Content-Type charset


andrewgaul:
S3Proxy responses include a charset in their Content-Type unlike Amazon S3:

```
application/xml; charset=UTF-8
```

Further some responses inconsistently return:

```
application/xml; charset=ISO-8859-1
```

PR #269, opened 2018-04-13
Add `UTF-8` character encoding to all S3 responses

randomcoder:
re #237 This makes all responses specify a character encoding of `UTF-8`. The cause of the inconsistent behaviour was jetty's configuration which uses `charset=ISO-8859-1` as its default if none other is given and jetty does not include a default character encoding for `application/xml`.

As jetty will add the encoding if it is not present, forcing `UTF-8` seems the most sensible solution and all the AWS xml responses specify an XML encoding of `UTF-8` in the xml preamble `<?xml version="1.0" encoding="UTF-8"?>`.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -182,6 +182,7 @@ public class S3ProxyHandler {
             "log-delivery-write"
     );
     private static final String XML_CONTENT_TYPE = "application/xml";
+    private static final String UTF_8 = "UTF-8";
     /** URLEncoder escapes / which we do not want. */
     private static final Escaper urlEscaper = new PercentEscaper("*-./_", true);
     @SuppressWarnings("deprecation")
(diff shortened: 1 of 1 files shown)


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -1315,13 +1321,12 @@ private void handleBlobList(HttpServletRequest request,
         }
         options.maxResults(maxKeys);
 
-        response.setCharacterEncoding("UTF-8");
-
         PageSet<? extends StorageMetadata> set = blobStore.list(containerName,
                 options);
 
+        response.setCharacterEncoding(UTF_8);
```
gaul: Should we set the character encoding next to the content type for consistency?

randomcoder: Sadly no, the encoding is permanently set once the writer is generated, with any `setCharacterEncoding` calls effectively ignored, so I had to set it immediately before instead. Caused me to scratch my head for a while did that.

gaul: Thanks for investigating this!  This makes sense that the `Writer` would get some state set during construction.

gaul:
Thank you for your contribution @randomcoder!

Issue #270, opened 2018-04-13
Unable to build S3Proxy with OpenJDK 9


timuralp:
When building with OpenJDK 9, I get the following stack trace from ErrorProne:
```
Caused by: java.lang.NullPointerException
    at com.sun.tools.javac.file.Locations.getPathEntries (Locations.java:157)
    at com.sun.tools.javac.file.Locations.getPathEntries (Locations.java:141)
    at com.sun.tools.javac.file.Locations.access$000 (Locations.java:80)
    at com.sun.tools.javac.file.Locations$BootClassPathLocationHandler.computePath (Locations.java:649)
    at com.sun.tools.javac.file.Locations$BootClassPathLocationHandler.lazy (Locations.java:677)
    at com.sun.tools.javac.file.Locations$BootClassPathLocationHandler.isDefault (Locations.java:566)
    at com.sun.tools.javac.file.Locations.isDefaultBootClassPath (Locations.java:115)
    at com.sun.tools.javac.util.BaseFileManager.isDefaultBootClassPath (BaseFileManager.java:137)
    at com.sun.tools.javac.main.Arguments.validate (Arguments.java:411)
    at com.sun.tools.javac.main.Main.compile (Main.java:215)
    at com.google.errorprone.ErrorProneCompiler.run (ErrorProneCompiler.java:249)
    at com.google.errorprone.ErrorProneCompiler.run (ErrorProneCompiler.java:159)
    at com.google.errorprone.ErrorProneCompiler.compile (ErrorProneCompiler.java:88)
```

I cannot reproduce this with OpenJDK 8. Upgrading the `error_prone_core` dependency to 2.1.2 resolves this for me. The JDK version is: 9.0.4+12-4 from Debian.

timuralp:
Although, building with updated `error_prone_core` results in 23 warnings. I'll try to submit a PR for updating it and fixing the warnings.

gaul:
Compile-time: error-prone packages javac internally which may cause these issues.  S3Proxy could add a Maven profile for JDK >= 9 which disables error-prone.  However, jclouds 2.2.0 may start requiring Java 8 which will allow S3Proxy to upgrade its problematic dependencies.

Run-time: does S3Proxy run with JDK 9?  I suspect that Guice issues may prevent this which S3Proxy could upgrade its dependency for.

gaul:
I verified that Java 9 can compile S3Proxy when disabling error-prone.  I opened a pull request which should address Java 9 run-time errors.

[...]

PR #272, opened 2018-04-14
Add jaxb-api as a dependency

gaul:
This allows Java 9 to run S3Proxy.  Java 9 removed `javax.xml` which
previous Java versions included so adding the explicit dependency
allows both versions to work.  References #270.


gaul:
Opened [JCLOUDS-1408](https://issues.apache.org/jira/browse/JCLOUDS-1408) for a similar, longer-term solution.

PR #273, opened 2018-04-17
shading guava/jclouds

patduin:
Hi :)
Following discussions here: https://issues.apache.org/jira/browse/JCLOUDS-1225 and here: https://github.com/gaul/s3proxy/issues/209
I made this PR that shades jclouds and shades/relocates guava. Our issue, as @massdosage described in JCLOUDS-1225, is that we got a dependency on hadoop which is on a rather old guava version conflicting with what s3proxy pulls in.

Following the Jetty shading I've went for a similar solution with guava. I'll add some inline comments with explanations.

Happy to have a discussion or make any changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,10 +156,26 @@
               <goal>shade</goal>
             </goals>
             <configuration>
-             <createDependencyReducedPom>false</createDependencyReducedPom>
-             <artifactSet>
+              <createDependencyReducedPom>true</createDependencyReducedPom>
+              <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                  <resource>META-INF/services/org.jclouds.providers.ProviderMetadata</resource>
+                </transformer>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                  <resource>META-INF/services/org.jclouds.apis.ApiMetadata</resource>
+                </transformer>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                  <resource>META-INF/services/org.jclouds.logging.config.LoggingModule</resource>
+                </transformer>
+              </transformers>
+              <artifactSet>
                 <includes>
                   <include>org.eclipse.jetty:*</include>
+                  <include>com.google.guava:guava</include>
+                  <include>org.apache.jclouds:*</include>
+                  <include>org.apache.jclouds.api:*</include>
+                  <include>org.apache.jclouds.provider:*</include>
                 </includes>
               </artifactSet>
               <relocations>
@@ -167,11 +183,15 @@
                   <pattern>org.eclipse.jetty</pattern>
                   <shadedPattern>${shade.prefix}.org.eclipse.jetty</shadedPattern>
                 </relocation>
+                <relocation>
+                  <pattern>com.google.common</pattern>
+                  <shadedPattern>${shade.prefix}.com.google.common</shadedPattern>
+                </relocation>
               </relocations>
             </configuration>
           </execution>
         </executions>
-      </plugin>      
+      </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-assembly-plugin</artifactId>
@@ -364,6 +384,11 @@
       <version>3.0.2</version>
     </dependency>
     <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>18.0</version>
+    </dependency>
+    <dependency>
       <groupId>commons-fileupload</groupId>
       <artifactId>commons-fileupload</artifactId>
       <version>1.3.3</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -156,22 +156,42 @@
               <goal>shade</goal>
             </goals>
             <configuration>
-             <createDependencyReducedPom>false</createDependencyReducedPom>
-             <artifactSet>
+              <createDependencyReducedPom>true</createDependencyReducedPom>
```
patduin: I think this was an oversight when it was added but it is better to use the dependency reduced pom as the classes we shade are now in the s3Proxy jar so no need to pull them in via the pom dependencies.
The `promoteTransitiveDependencies` makes sure and transitive deps are promoted and added to the dependency reduced pom.

pom.xml
```diff
@@ -156,22 +156,42 @@
...
+                </transformer>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                  <resource>META-INF/services/org.jclouds.logging.config.LoggingModule</resource>
+                </transformer>
+              </transformers>
+              <artifactSet>
                 <includes>
                   <include>org.eclipse.jetty:*</include>
+                  <include>com.google.guava:guava</include>
+                  <include>org.apache.jclouds:*</include>
```
patduin: No need to relocate jclouds just needs to be shaded so it picks up the guava relocations

pom.xml
```diff
@@ -373,6 +398,21 @@
       <artifactId>jclouds-allblobstore</artifactId>
       <version>${jclouds.version}</version>
     </dependency>
+    <dependency>
```
patduin: had to add these otherwise I can't reference them in the shade plugin

pom.xml
```diff
@@ -363,6 +383,11 @@
       <artifactId>jsr305</artifactId>
       <version>3.0.2</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>18.0</version>
+    </dependency>
```
gaul: Most of this pull request makes sense except that we are trading one set of user problems with Guava 21 for another with Guava 18.  Could S3Proxy use the shaded jclouds Guava imports instead?  It looks like S3Proxy could mostly eliminate its use of Guava as well.

pom.xml
gaul: This section confuses me -- what is going on here?

patduin: the jclouds dependences were pulled in transitively, the maven shade plugin doesn't pick them up, so I had to add them explicitly

pom.xml
patduin: Guava was added explicitly because the shade plugin doesn't pickup transitive dependencies. I went for 18.0 because that is the version jclouds-2.1.0 was pulling in. We could change this. 
S3Proxy will compile against this version and during the shading the classes will be updated to use the relocated guava classes. 
Any module using this will not see any guava dependency and can define their own.

patduin:
about java 9 modules: good question I haven't got any experience using java 9 modules so I would not be able to say for sure.

massdosage:
Given that our use case for this needs to support Java 8 and 7, Java 9's modularity features won't be able to help us (I'm not even sure it can as things currently stand).

pom.xml
gaul: Should this include the the entire tree of dependencies, e.g., b2, google-cloud-storage, openstack-swift, etc.?

pom.xml
gaul: Thanks for explaining!  I looked at S3Proxy use of Guava and believe that we should excise it from the public interfaces at least.  I only found a few uses of `Optional` parameters which I replaced with `@Nullable` arguments in ec00aeb8cd23468366ed1a0eeaa9bae4f3bad6eb.

gaul:
> Given that our use case for this needs to support Java 8 and 7, Java 9's modularity features won't be able to help us (I'm not even sure it can as things currently stand).

Agree that Java 9 cannot help your use case.  I will caution that jclouds will likely require Java 8 soon and S3Proxy will follow.

pom.xml
patduin: :) that is a good question and it depends. If those jars use guava they need to be added to the list of shading deps and added to the pom. If they don't and I suspect we're ok for it, they don't need to be added.

Just to clarify I've only added the jcloud jars so I can reference them in the mvn shade plugin:
`<include>org.apache.jclouds:*</include>`

patduin:
Please let me know if there is anything you'd like me to change :)

pom.xml
massdosage: Just doing jclouds resolves our downstream issue. I guess we could start with just this and expand it later as/when necessary?

pom.xml
massdosage: @gaul sounds good. I think we still need the above otherwise the shading won't work, unless we wanted to move the shading and relocation up into jclouds itself...

massdosage:
@gaul Could you please let us know if you think this is going to make it into S3Proxy and if so a vague timeline for when we could expect a release? We're blocked on this so if it's unlikely or going to take a long time we need to move on to a Plan B on our side. Thanks!

patduin:
We've shaded s3proxy in our project working around the problem, I'm gonna close this. Thanks for the feedback and time.

Issue #202, opened 2017-03-14
InvalidLocationConstraint when creating bucket


kishore25kumar:
We are using s3proxy to connect to azure storage. When we are creating a bucket using nodejs sdk I am getting InvalidLocationConstraint error. Here is the stack trace

```
{ InvalidLocationConstraint: The specified location constraint is not valid. For more information about Regions, see How to Select a Region for Your Buckets.
    at Request.extractError (/Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/services/s3.js:538:35)
    at Request.callListeners (/Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/sequential_executor.js:105:20)
    at Request.emit (/Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/request.js:668:14)
    at Request.transition (/Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/request.js:670:12)
    at Request.callListeners (/Users/battula/eloise/foundational-services/node-s3proxy/node_modules/aws-sdk/lib/sequential_executor.js:115:18)
  message: 'The specified location constraint is not valid. For more information about Regions, see How to Select a Region for Your Buckets.',
  code: 'InvalidLocationConstraint',
  region: null,
  time: 2017-03-14T09:40:15.509Z,
  requestId: null,
  extendedRequestId: undefined,
  cfId: undefined,
  statusCode: 400,
  retryable: false,
  retryDelay: 75.30141439047937 }
```
If I use java sdk it works fine. After a little debugging I found that if I pass region as azureblob while creating the bucket then it is working fine.

[...]

PR #277, opened 2018-05-29
Allow setting property jclouds.regions via env variables when running in docker

pscanf:
Also, default to`jcloud.regions=us-east-1` to avoid issues with the nodejs aws sdk like the one described in #202.


gaul:
Thank you for your contribution @pscanf!

PR #280, opened 2018-07-23
Update README.md

kartenkarsten:
updated required java version to java 8

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Usage without Docker
 --------------------
 Users can [download releases](https://github.com/gaul/s3proxy/releases)
 from GitHub.  Developers can build the project by running `mvn package` which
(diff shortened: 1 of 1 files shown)


gaul:
Which part of S3Proxy requires Java 7 to run?

kartenkarsten:
> Which part of S3Proxy requires Java 7 to run?
as far as I now are all parts java 8 compatible but in readme was still written requires java 7

gaul:
I see the confusion -- can you amend the pull request to say that S3Proxy requires Java 7 or later?

kartenkarsten:
I didn't tried java 9 is it already fully supported?

gaul:
I verified compatibility with Java 8 and 11.

PR #286, opened 2018-11-14
Keystone v3 support with properties

reimannf:
As already documented in the Wiki https://github.com/gaul/s3proxy/wiki/Storage-backend-examples#swift-keystone-v30, you can use a Swift backend with Keystone v3 authentification.
This PR makes the needed `jcloud` configuration options available in the Dockerfile and pass the environment as properties to the java runtime.


gaul:
Thank you for your contribution @reimannf!

PR #293, opened 2019-02-09
Add partial support for list objects v2

gaul:
Not supporting `fetch-owner` until jclouds adds support for this.
Several applications like AWS CLI now require this RPC.


gaul:
@jumenzel I believe you had interest in adding this to jclouds and S3Proxy.  Could you review this?

Issue #81, opened 2015-10-06
swift: how to select region? It is always the first entry selected.


Dexus:
```
main o.j.l.s.i.GetRegionIdMatchingProviderURIOrNull:74 |::] failed to find key for value  https://auth.cloud.ovh.net/v2.0 in {GRA1=https://storage.gra1.cloud.ovh.net/v1/AUTH_xxx, BHS1=https://storage.bhs1.cloud.ovh.net/v1/AUTH_xxx, SBG1=https://storage.sbg1.cloud.ovh.net/v1/AUTH_xxx}; choosing first: GRA1
```
my config:
```
s3proxy.endpoint=http://127.0.0.1:8080
s3proxy.authorization=aws-v2
s3proxy.identity=xxx
s3proxy.credential=xxx
jclouds.provider=openstack-swift
jclouds.endpoint=https://auth.cloud.ovh.net/v2.0
jclouds.api=swift
jclouds.regions=SBG1
jclouds.identity=1xx9:xxx
jclouds.credential=xxx
```

andrewgaul:
@Dexus S3Proxy previously lacked support for specifying regions for Swift-based object stores.  I added support and an example for Rackspace Cloud Files here:

https://github.com/andrewgaul/s3proxy/wiki/Provider-examples#rackspace-cloud-files

In the current implementation, you specify a single `jclouds.region` (note not plural).  Note that Swift has multiple ways of specifying regions so you may need to experiment with different endpoints.

PolSpock:
Hello,
I have the same problem :/

```
./s3proxy --properties test.conf 
[s3proxy] W 02-14 14:53:56.630 main o.j.l.s.i.GetRegionIdMatchingProviderURIOrNull:74 |::] failed to find key for value https://auth.cloud.ovh.net/v3/ in {DE1=https://storage.de1.cloud.ovh.net/v1/AUTH_XXXX, UK1=https://storage.uk1.cloud.ovh.net/v1/AUTH_XXXX, SBG3=https://storage.sbg3.cloud.ovh.net/v1/AUTH_XXXX, WAW1=https://storage.waw1.cloud.ovh.net/v1/AUTH_XXXX, BHS3=https://storage.bhs3.cloud.ovh.net/v1/AUTH_XXXX, GRA3=https://storage.gra3.cloud.ovh.net/v1/AUTH_XXXX}; choosing first: DE1
```

OVH is mutli region
Someone can help me ? 😄 

Regards,

[...]

PR #295, opened 2019-03-05
Fix for OVH swift storage (url mismatch for region)

Ogrodniczek:
I've tried to use the proxy for OVH swift storage and got the same problem as #81 . The main problem was that `BlobStore blobStore = context.getBlobStore();` tries to compare different endpoints basing on values of the dictionary returned by the OVH API and the JCLOUDS_ENDPOINT variable. It mismatches hardly and it's falling back to the first region, that is DE1.

`[s3proxy] W 03-05 23:32:53.312 main o.j.l.s.i.GetRegionIdMatchingProviderURIOrNull:74 |::] failed to find key for value https://auth.cloud.ovh.net/v3/ in {DE1=https://storage.de1.cloud.ovh.net/v1/AUTH_d443, UK1=https://storage.uk1.cloud.ovh.net/v1/AUTH_d443, WAW1=https://storage.waw1.cloud.ovh.net/v1/AUTH_d443, GRA5=https://storage.gra5.cloud.ovh.net/v1/AUTH_d443, BHS3=https://storage.bhs3.cloud.ovh.net/v1/AUTH_d443, SBG5=https://storage.sbg5.cloud.ovh.net/v1/AUTH_d443}; choosing first: DE1
`

I've managed to tune the code a little bit and it happened to fix my problem, now I can define proper region in JCLOUDS_REGION variable. (JCLOUDS_REGIONS also has to be filled and has to contain the chosen region).
Also it seem that all tests are passing, therefore, here is the PR.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/Main.java
+++ b/src/main/java/org/gaul/s3proxy/Main.java
@@ -310,13 +310,13 @@ public final class Main {
         }
 
         BlobStoreContext context = builder.build(BlobStoreContext.class);
-	BlobStore blobStore;
-	if (region != null) {
-	        blobStore = ((RegionScopedBlobStoreContext) context)
+        BlobStore blobStore;
+        if (region != null) {
+            blobStore = ((RegionScopedBlobStoreContext) context)
                     .getBlobStore(region);
-	} else {
-		blobStore = context.getBlobStore();
-	}
+        } else {
+            blobStore = context.getBlobStore();
+        }
         return blobStore;
     }
 
--- a/src/main/java/org/gaul/s3proxy/Main.java
+++ b/src/main/java/org/gaul/s3proxy/Main.java
@@ -311,7 +311,8 @@ public final class Main {
 
         BlobStoreContext context = builder.build(BlobStoreContext.class);
         BlobStore blobStore;
-        if (region != null) {
+        if (context instanceof RegionScopedBlobStoreContext &&
+                region != null) {
             blobStore = ((RegionScopedBlobStoreContext) context)
                     .getBlobStore(region);
         } else {
--- a/src/main/java/org/gaul/s3proxy/Main.java
+++ b/src/main/java/org/gaul/s3proxy/Main.java
@@ -310,12 +310,13 @@ public final class Main {
         }
 
         BlobStoreContext context = builder.build(BlobStoreContext.class);
-        BlobStore blobStore = context.getBlobStore();
-        if (context instanceof RegionScopedBlobStoreContext &&
-                region != null) {
-            blobStore = ((RegionScopedBlobStoreContext) context)
+	BlobStore blobStore;
+	if (region != null) {
+	        blobStore = ((RegionScopedBlobStoreContext) context)
                     .getBlobStore(region);
-        }
+	} else {
+		blobStore = context.getBlobStore();
+	}
         return blobStore;
     }
 


src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -310,11 +310,12 @@ private static BlobStore createBlobStore(Properties properties,
         }
 
         BlobStoreContext context = builder.build(BlobStoreContext.class);
-        BlobStore blobStore = context.getBlobStore();
-        if (context instanceof RegionScopedBlobStoreContext &&
-                region != null) {
+        BlobStore blobStore;
+        if (region != null) {
```
gaul: How does this differ from the previous code?  It seems unsafe since it no longer checks whether `context` is actually a `RegionScopedBlobStoreContext`.  Also I think OVH is Swift and thus should be a `RegionScopedBlobStoreContext`?

src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -310,11 +310,13 @@ private static BlobStore createBlobStore(Properties properties,
         }
 
         BlobStoreContext context = builder.build(BlobStoreContext.class);
-        BlobStore blobStore = context.getBlobStore();
+        BlobStore blobStore;
```
gaul: I don't oppose this change but how does it address your symptoms?  Wouldn't it have the same effect either way?

Ogrodniczek: So, I'd say that OVH swift misbehave when getBlobStore is called without any argument. But really I do not know internals of swift, nor jclouds.openstack.swift.
Here is a stacktrace from current latest docker image:

`[s3proxy] W 03-06 10:26:48.936 main o.j.l.s.i.GetRegionIdMatchingProviderURIOrNull:74 |::] failed to find key for value https://auth.cloud.ovh.net/v3 in {DE1=https://storage.de1.cloud.ovh.net/v1/AUTH_d446...43, UK1=https://storage.uk1.cloud.ovh.net/v1/AUTH_d446...43, WAW1=https://storage.waw1.cloud.ovh.net/v1/AUTH_d446...43, GRA5=https://storage.gra5.cloud.ovh.net/v1/AUTH_d446...43, BHS3=https://storage.bhs3.cloud.ovh.net/v1/AUTH_d446...43, SBG5=https://storage.sbg5.cloud.ovh.net/v1/AUTH_d446...43}; choosing first: DE1
[s3proxy] E 03-06 10:26:48.941 main org.gaul.s3proxy.Main:236 |::] Exception in thread "main" 
[s3proxy] E 03-06 10:26:48.941 main org.gaul.s3proxy.Main:236 |::] java.lang.IllegalArgumentException: region DE1 not in [UK1]
[s3proxy] E 03-06 10:26:48.942 main org.gaul.s3proxy.Main:236 |::] 	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:145)
[s3proxy] E 03-06 10:26:48.942 main org.gaul.s3proxy.Main:236 |::] 	at org.jclouds.openstack.swift.v1.blobstore.RegionScopedBlobStoreContext.checkRegionId(RegionScopedBlobStoreContext.java:88)
[s3proxy] E 03-06 10:26:48.942 main org.gaul.s3proxy.Main:236 |::] 	at org.jclouds.openstack.swift.v1.blobstore.RegionScopedBlobStoreContext.getBlobStore(RegionScopedBlobStoreContext.java:72)
[s3proxy] E 03-06 10:26:48.942 main org.gaul.s3proxy.Main:236 |::] 	at org.jclouds.openstack.swift.v1.blobstore.RegionScopedBlobStoreContext.getBlobStore(RegionScopedBlobStoreContext.java:122)
[s3proxy] E 03-06 10:26:48.942 main org.gaul.s3proxy.Main:236 |::] 	at org.gaul.s3proxy.Main.createBlobStore(Main.java:313)
[s3proxy] E 03-06 10:26:48.943 main org.gaul.s3proxy.Main:236 |::] 	at org.gaul.s3proxy.Main.main(Main.java:112)`

>Wouldn't it have the same effect either way?
After my change there is no error, region is chosen properly, so it works differently.

gaul: Now I understand -- calling `context.getBlobStore` has a side-effect.

PR #296, opened 2019-04-23
Add middleware for AES-encryption of payload

FKatenbrink:
This PR introduces a middleware for client-side encryption of the payload with AES-CTR as requested in https://github.com/gaul/s3proxy/issues/126.

The 16 byte initialization vector (IV) and 1 byte size of the IV is stored in the beginning of the payload. This results in a 17 byte difference of the encrypted payload size and the plain text payload size. An alternative could be to store the IV in the user metadata of the payload.

We can configure the following properties to enable the encryption:

```
s3proxy.encrypted-blobstore=true
s3proxy.encrypted-blobstore-key=<SECRET>
s3proxy.encrypted-blobstore-salt=<SALT>
```


gaul:
This pull request looks good!  It needs some wiki documentation for other users but you can do this separately and add a link to the README in a subsequent pull request.

Before merging, I would like to explore if there are other approaches that may provide compatibility with other applications doing client-side encryption.  Also, how should S3Proxy handle buckets with mixed encrypted and unencrypted objects?  If instead the implementation stored user metadata as you suggest, S3Proxy could dynamically decrypt blobs.  This could store the encryption type `AES/CTR/PKCS5Padding` in one key and the IV in another so the blobs have the same size.  What do you think?

elemental-lf:
@gaul I'm not using `s3proxy` at the moment and so can't justify investing time into it right now. I'm sorry. One last note: I think the referenced Java example isn't applicable here as it uses the AWS SDK. The logic would need to be re-implemented. Or it might be possible to use some low-level classes from the AWS SDK to just do the encrypted object handling and then pass it on to `jclouds`, I haven't looked into it.

FKatenbrink:
> @gaul I'm not using `s3proxy` at the moment and so can't justify investing time into it right now. 

Same for me. Feel free to adapt the PR to the format suggested by @elemental-lf.

benweet:
@FKatenbrink why not use AES/GCM instead to ensure data integrity?

gaul:
Superseded by #398.

Issue #304, opened 2019-06-24
s3proxy does not reject invalid range headers in UploadPartCopy requests.


timuralp:
When submitting an UploadPartCopy request to S3Proxy with an invalid source range, the full object is used as the source, as opposed to returning an error. AWS S3 returns the following error message in that case:
```
<Error>
    <Code>InvalidArgument</Code>
    <Message>The x-amz-copy-source-range value must be of the form bytes=first-last
        where first and last are the zero-based offsets of the first and last bytes to copy
    </Message>
    <ArgumentName>x-amz-copy-source-range</ArgumentName>
    <ArgumentValue>0-1024</ArgumentValue>
    <RequestId>[...]</RequestId>
    <HostId>[...]</HostId>
</Error>
```

S3Proxy should return a similar error after validating that the range is specified in the expected format (`bytes=%d-%d`).

PR #305, opened 2019-06-24
UploadPartCopy: Fail if improper source range.

timuralp:
S3Proxy should check the format of the x-amz-copy-source-range value
when handling an UploadPartCopy request. The patch changes the behavior
to return an InvalidArgument error along with the matching error
description to AWS S3. To match the error format, the S3Exception class
is modified to return the actual error message, rather than the String
version of the error code.

Corresponding tests will be added to ceph/s3-tests.

Fixes: #304

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3Exception.java
+++ b/src/main/java/org/gaul/s3proxy/S3Exception.java
@@ -63,7 +63,7 @@ public final class S3Exception extends Exception {
 
     @Override
     public String getMessage() {
-        StringBuilder builder = new StringBuilder().append(error);
+        StringBuilder builder = new StringBuilder().append(super.getMessage());
         if (!elements.isEmpty()) {
             builder.append(" ").append(elements);
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/gaul/s3proxy/S3Exception.java
```diff
@@ -63,7 +65,12 @@ S3ErrorCode getError() {
 
     @Override
     public String getMessage() {
-        StringBuilder builder = new StringBuilder().append(error);
+        StringBuilder builder;
+        if (message != null) {
+            builder = new StringBuilder().append(message);
```
gaul: Could you assign `StringBuilder` only once above and mutate it with `append(message)`, possibly with the ternary `?:` operator?

src/main/java/org/gaul/s3proxy/S3Exception.java
```diff
@@ -50,6 +51,7 @@
                 Map<String, String> elements) {
         super(requireNonNull(message), cause);
         this.error = requireNonNull(error);
+        this.message = message;
```
gaul: Can you retrieve this via `super.getMessage`?

src/main/java/org/gaul/s3proxy/S3Exception.java
timuralp: I realized this simplifies a bit more: `message` must also be set and there is a corresponding `requireNonNull` check. Done.

src/main/java/org/gaul/s3proxy/S3Exception.java
timuralp: Done

gaul:
Thank you for your contribution @timuralp!

PR #313, opened 2019-09-24
Removed container root user requirement

tux-o-matic:
The service will now listen on port 8080 with a regular UID.
The local /data volume can also be written to by this UID.

Tested in OpenShift.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,7 @@ ENV \
 
 EXPOSE 8080
 VOLUME /data
+RUN chmod 775 /data
 USER 1001
 
 ENTRYPOINT ["/opt/s3proxy/run-docker-container.sh"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -39,7 +39,9 @@ ENV \
     JCLOUDS_KEYSTONE_SCOPE="" \
     JCLOUDS_KEYSTONE_PROJECT_DOMAIN_NAME=""
 
-EXPOSE 80
+EXPOSE 8080
```
gaul: Won't this break the configuration for any users currently relying on port 80?

tux-o-matic: That's just the internal service, Docker users currently depending on the service being available on port 80 just have to map 80:8080. So if this PR is merged, a little update to the README would be a good idea.

gaul: Lots of people deploy S3Proxy in a variety of CI environments.  I don't want to break anyone's existing configuration.  Is there any way to transition cleanly?

tux-o-matic: It's possible to declare 'EXPOSE' for 80 and 8080 in the Dockerfile to document the options and then via an ENV variable tell the java process which port to listen on. That way 80 could remain the default to prevent a breaking change.

gaul:
Is there a path forward with this pull request?

Issue #331, opened 2020-09-08
Multipart upload CORS headers


Renkas:
Currently multipart upload does not fully support CORS headers.

Find more details from older issue: https://github.com/gaul/s3proxy/issues/142#issuecomment-513499929

@reimannf seems to have a solution for this. Could you please implement it?

gaul:
@Renkas Could you test the latest master?

PR #332, opened 2020-09-10
CORS header for MultiPart Requests

reimannf:
Adding CORS headers `Access-Control-Allow-Origin` and `Access-Control-Allow-Methods` in Multipart Reposnes if the
`Origin` Header is included in the Request and does match the CORS rules.
This was requested in #331.

Unfortunately my spare time is very limited at the moment and I was not able to write a test case for the multi part uploads. I tried, but there seems to be no easy way to retrieve the response Headers e.g. from `InitiateMultipartUploadResult` like here: https://github.com/gaul/s3proxy/blob/master/src/test/java/org/gaul/s3proxy/AwsSdkTest.java#L481-L482 to verify that CORS headers are present.

So maybe some else could step in here and add a test.

I am sorry that I cannot provide more in the moment.


gaul (changes requested):
This seems reasonable.  Could you rebase and squash this PR?

reimannf:
Clean history in #334

PR #339, opened 2020-12-27
support custom s3 proxy endpoint in docker

johnnyaug:


gaul:
Thank you for your contribution @johnnyaug!

johnnyaug:
Thanks @gaul, could you please re-trigger the build so that a new image is published?

gaul:
I retriggered and believe that this succeeded.  Docker Hub seems to be some new kind of rate limiting which this project randomly encounters due to lacking an API token.  The number of layers in the underlying Docker images and pulling twice as many because of multistage may also influence this.

PR #341, opened 2021-01-04
add virtual host to docker

johnnyaug:

PR #342, opened 2021-01-09
Use the old environment variable for virtual host

CurosMJ:
This project used the env var `S3PROXY_VIRTUALHOST`, but then this PR: #341 added another declaration that used `S3PROXY_VIRTUAL_HOST`.

This PR fixes this, and goes back to `S3PROXY_VIRTUALHOST`.

Thanks !


gaul:
@johnnyaug could you comment on this?

johnnyaug:
@CurosMJ is right. I didn't notice that the command was already using a variable for virual host, so I introduced a new one instead of using the existing one. This PR should be merged. Sorry for the mixup

gaul:
Thank you for your contribution @CurosMJ!

PR #344, opened 2021-01-25
Better support for caching scenarios of CORS responses

### LINKED ISSUE #343: Missing CORS header in the response
Hello,

First, thank you for the project, it is the only S3 proxy available that we found that support a progress bar while downloading, which help us!

We setup the s3proxy the following way
```
    - S3PROXY_AUTHORIZATION=none
    - S3PROXY_CORS_ALLOW_ALL=true
    - S3PROXY_CORS_ALLOW_ORIGINS=*
    - S3PROXY_CORS_ALLOW_METHODS=GET PUT POST HEAD DELETE
    - S3PROXY_CORS_ALLOW_HEADERS=*
    - JCLOUDS_PROVIDER=s3
```

As far as I understand CORS should allow any origins with any headers and any methods.

Unfortunately, all the CORS header are missing from the response

Here is the curl command we use:
`curl  -H "Access-Control-Request-Method: GET" -H "Origin: http://localhost" --head http://localhost/<bucket_name>/open-sans-3.ttf`
 
And the response we got:
```
Date: Wed, 13 Jan 2021 09:14:19 GMT
x-amz-request-id: XXXXXXXXXXXXXXX
Content-Type: binary/octet-stream
ETag: "XXXXXXXXXXXXXXX"
Last-Modified: Tue, 29 Dec 2020 16:30:17 GMT
x-amz-storage-class: STANDARD
Content-Length: 28100
Server: Jetty(9.2.z-SNAPSHOT)
```

I can confirm that when we are curling directly the bucket with the same command, the following header will be in the response:

```
access-control-allow-origin: *
access-control-allow-methods: GET, HEAD, DELETE, PUT, POST
access-control-expose-headers: Access-Control-Allow-Origin
```

I am confused if we are doing something wrong, if it's a bug or a missing feature at the moment.

reimannf:
If the allowed origin `*` is configured e.g. through config option `s3proxy.cors-allow-all=true` the `ACCESS_CONTROL_ALLOW_ORIGIN` header in CORS responses will now always be `*` and not the requested host from the `Origin` header. If only concrete allowed origins are configured the behaviour remains unchanged.
Furthermore the `ACCESS_CONTROL_ALLOW_METHODS` header will always include all allowed methods in the CORS response.
This is inline with a native S3 CORS response and will help esp. in CDN caching scenarios. All requested Origins can be served with a cached response.

Further changes:
* S3Proxy only accepts allowed methods from `CrossOriginResourceSharing.SUPPORTED_METHODS`, preventing operators to configure methods which do not provide CORS headers atm, like `HEAD` or `DELETE`.
* If configured allowed origins contains `*` all other configured origins are skipped and are not pattern matched -> Allow any origin
* Testcase for `cors-allow-all=true` added

Closes #343

### DIFF
--- a/src/main/java/org/gaul/s3proxy/CrossOriginResourceSharing.java
+++ b/src/main/java/org/gaul/s3proxy/CrossOriginResourceSharing.java
@@ -35,9 +35,10 @@ import org.slf4j.LoggerFactory;
 
 final class CrossOriginResourceSharing {
     public static final Collection<String> SUPPORTED_METHODS =
-            ImmutableList.of("GET", "PUT", "POST", "HEAD");
+            ImmutableList.of("GET", "PUT", "POST");
 
     private static final String HEADER_VALUE_SEPARATOR = ", ";
+    private static final String ALLOW_ANY_ORIGIN = "*";
     private static final String ALLOW_ANY_HEADER = "*";
 
     private static final Logger logger = LoggerFactory.getLogger(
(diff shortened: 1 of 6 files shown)


gaul:
Thank you for your contribution @reimannf!  I appreciate you handling all these CORS issues and including tests.

tatiesmars:
Nice! We will test and let u know! Appreciate the help 👍

PR #345, opened 2021-03-04
Make CrossOriginResourceSharing public

snoe925:
I think cannot construct a S3ProxyHandler which is public without this class being public to pass to the constructor.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/CrossOriginResourceSharing.java
+++ b/src/main/java/org/gaul/s3proxy/CrossOriginResourceSharing.java
@@ -33,7 +33,7 @@ import com.google.common.collect.Lists;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-final class CrossOriginResourceSharing {
+public class CrossOriginResourceSharing {
     protected static final Collection<String> SUPPORTED_METHODS =
             ImmutableList.of("GET", "PUT", "POST");
 


src/main/java/org/gaul/s3proxy/CrossOriginResourceSharing.java
```diff
@@ -33,7 +33,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-final class CrossOriginResourceSharing {
+public class CrossOriginResourceSharing {
```
gaul, line 36: Why remove `final`?  This causes Checkstyle failures.  Also don't you need the constructors to be `public`?

gaul (changes requested):
I am curious about how you extend S3Proxy.  Would you mind sharing some details of your use case?

snoe925:
The use case was the same as the comment here:  https://github.com/gaul/s3proxy/blob/master/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java#L113 "HTTP server-independent handler for S3 requests."
I have a modified / old / different server implementation.  It's easier on me to have classes that don't reference Jetty.  Not for any particularly sound engineering reasoning.
I am also ok with "*" CORS all as the server is internal.  A better change might be to have it be @Nullable, and then construct the default CrossOriginResourceSharing in the constructor of S3ProxyHandler.
To address the protected constructors, I constructed the instance with a subclass as follows.
    class MyCrossOriginResourceSharing extends CrossOriginResourceSharing {}
but looking back, that was obtuse.

gaul:
> I am also ok with "*" CORS all as the server is internal. A better change might be to have it be @nullable, and then construct the default CrossOriginResourceSharing in the constructor of S3ProxyHandler.

I think it is reasonable for the `CrossOriginResourceSharing` constructors to be `public` and the `S3ProxyHandler` parameter to be `@nullable`.  Do you plan to send another pull request?

> I have a modified / old / different server implementation. It's easier on me to have classes that don't reference Jetty. Not for any particularly sound engineering reasoning.

I am happy to accept PRs that make embedding easier.  Most people are happy with Jetty but if there is a better way to plug in alternatives or some other approach please let me know.

snoe925:
I am about 1/2 way into a real integration with the @Nullable style for default CORS handling.
I'll have a new pull request on code that I've really tested.
I also figured out my maven issues so that I can checkstyle locally.

gaul:
Do you need any help with this?  I would like to publish S3Proxy 1.8.0.  Would it be sufficient to make the class and constructors `public`?

snoe925:
I made a fresh pull.  I ran into more checkstyle errors because of the finals when I tried public and public constructors.
Note I also had to add HEAD to get it really working with a real client.

PR #346, opened 2021-03-06
Make CORS handling optional

snoe925:
This makes it possible to construct the S3ProxyHandler.  
Making the CrossOriginResourceSharing public with public constructors caused more CheckStyle errors because of all the finals.
In my testing with AWS client v2 I also had to add HEAD to the methods.


gaul:
Thank you for your contribution @snoe925!

PR #347, opened 2021-03-15
Support for Junit 5 Extension mechanism

### LINKED ISSUE #288: Junit5: support jupiter engine
Are there plans to support `junit5` via the extension mechanism? Do you need support (I might be able to help, though I did not yet looked into details)

xgourmandin:
A proposal to close #288 

This feature adds support for Junit 5 test engine.

A user simply has to add this to its test class (as shown in the unit test added):
```
@RegisterExtension
static S3ProxyExtension EXTENSION = S3ProxyExtension.builder()
         .withCredentials("access", "secret")
         .build();
```


pom.xml
```diff
@@ -1,455 +1,468 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
gaul, line 1: Could you not reformat this entire file?

pom.xml
```diff
@@ -1,455 +1,468 @@
...
+                            <name>junit</name>
+                            <value>false</value>
+                        </property>
+                    </properties>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>com.github.spotbugs</groupId>
+                <artifactId>spotbugs-maven-plugin</artifactId>
+                <version>4.2.0</version>
```
gaul, line 254: Please do not include irrelevant changes.  I can merge these as individual PRs.

gaul (changes requested):
Not sure why Travis fails due to rate limits since it shouldn't be doing this except for the master branch.

@ThanksForAllTheFish, @ddcprg, @massdosage could you review this?  I am not familiar with JUnit 5.

massdosage:
@gaul sure, I'll hopefully have some time tomorrow or next week and will take a look. Agree on your comments above about formatting etc, makes it a bit hard to see what exactly has changed.

src/main/java/org/gaul/s3proxy/junit/S3ProxyExtension.java
```diff
@@ -0,0 +1,189 @@
...
+import org.jclouds.blobstore.BlobStoreContext;
+import org.jclouds.blobstore.domain.StorageMetadata;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public final class S3ProxyExtension
```
massdosage, line 40: There is a *lot* of repeated code between this and `S3ProxyRule`. I have worked on migrating quite a few projects to support JUnit4 Rules and JUnit5 extensions and this is how I have approached it:

* Move all the common, non-JUnit specific code out from the existing Rule into a new "Core" class.
* Have the Rule call out to this "Core" class from the various JUnit lifecycle methods.
* All existing unit tests and usages of the Rule shouldn't need to change.
* Implement the Extension in a similar manner to call out to the "Core" from the various JUnit5 lifecyle methods it implements.
* Add a test for the Extension along the same lines as the Rule test.
* Decide on which JUnit jars you want to mark with "compile", "test" or "provided" scopes in Maven depending how the project is used downstream and whether you want users to inherit JUnit dependencies transitively or whether you want them to explicitly pick and choose the JUnit versions they want.
* Update documentation describing how to use JUnit5 extension alongside the JUnit4 rule so users can use either.

massdosage (changes requested):

src/main/java/org/gaul/s3proxy/junit/S3ProxyExtension.java
xgourmandin: I'm totally ok with the code duplication comment.

I'm struggling with it because the JUnit 4 rule extends some Junit class and the JUnit 5 extension implements some JUnit interface.

I've already added a test for the extension (4th bullet)

massdosage: I think the only way you can refactor out the common code is using composition rather than inheritance since the JUnit4 Rule already extends another class like you said. Here is an example that might be helpful - [BeejuJunitExtension](https://github.com/HotelsDotCom/beeju/blob/main/src/main/java/com/hotels/beeju/extensions/BeejuJUnitExtension.java) and [BeejuJUnitRule](https://github.com/HotelsDotCom/beeju/blob/main/src/main/java/com/hotels/beeju/BeejuJUnitRule.java) both delegate most of the functionality to [BeejuCore](https://github.com/HotelsDotCom/beeju/blob/main/src/main/java/com/hotels/beeju/core/BeejuCore.java) (which in turn doesn't actually know anything about JUnit). The extension and the rule keep extending/implementing the same JUnit classes and interfaces, you just move the code which isn't dependent on a specific version of JUnit out into the core.

xgourmandin: Yes that's the way it shall be done.
Will do as soon as I can.

xgourmandin: I've done the factorization. I've kept the Builder in each class to avoid breaking existing JUnit 4 test classes.

src/main/java/org/gaul/s3proxy/junit/S3ProxyExtension.java
```diff
@@ -0,0 +1,108 @@
...
+
+        public Builder withCredentials(AuthenticationType authType,
+                                       String accessKey, String secretKey) {
+            builder.withCredentials(authType, accessKey, secretKey);
+            return this;
+        }
+
+        public Builder withCredentials(String accessKey, String secretKey) {
+            return withCredentials(AuthenticationType.AWS_V2_OR_V4, accessKey,
+                    secretKey);
```
massdosage, line 51:
```suggestion
          builder.withCredentials(accessKey, secretKey);
          return this;
```
I think this might be better to keep it consistent and to keep the logic of defaulting to `AuthenticationType.AWS_V2_OR_V4` in one place.

src/main/java/org/gaul/s3proxy/junit/S3ProxyExtension.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+/**
+ * A JUnit 5 Extension that manages an S3Proxy instance which tests
+ * can use as an S3 API endpoint.
+ */
+public final class S3ProxyExtension
+        implements AfterAllCallback, BeforeAllCallback {
```
massdosage, line 31: The JUnit4 rule extends `ExternalResource`  and then overrides `before()` which is the equivalent of `beforeEach` in JUnit5 not `beforeAll`. So to be consistent you need to change this to implement `BeforeEachCallback`. The same goes for `AfterEachCallback`.

src/main/java/org/gaul/s3proxy/junit/S3ProxyRule.java
```diff
@@ -16,26 +16,15 @@
...
-import org.gaul.s3proxy.S3Proxy;
-import org.jclouds.ContextBuilder;
-import org.jclouds.blobstore.BlobStore;
-import org.jclouds.blobstore.BlobStoreContext;
-import org.jclouds.blobstore.domain.StorageMetadata;
 import org.junit.rules.ExternalResource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.Beta;
```
massdosage, line 26: @gaul so we still want the "beta" annotation on the rule? If so (or even if not) should we apply it to the extension?

src/main/java/org/gaul/s3proxy/junit/S3ProxyRule.java
```diff
@@ -45,60 +34,44 @@
...
-            this.authType = authType;
-            this.accessKey = accessKey;
-            this.secretKey = secretKey;
+                                                        String accessKey, String secretKey) {
+            builder.withCredentials(authType, accessKey, secretKey);
             return this;
         }
 
         public Builder withCredentials(String accessKey, String secretKey) {
             return withCredentials(AuthenticationType.AWS_V2_OR_V4, accessKey,
```
massdosage, line 54:
```suggestion
          builder.withCredentials(accessKey, secretKey);
          return this;
```
(as per previous comment in the extension).

src/test/java/org/gaul/s3proxy/junit/S3ProxyExtensionTest.java
```diff
@@ -0,0 +1,93 @@
...
+    static final S3ProxyExtension EXTENSION = S3ProxyExtension
+            .builder()
+            .withCredentials("access", "secret")
+            .build();
+
+    private static final String MY_TEST_BUCKET = "my-test-bucket";
+
+    private static AmazonS3 s3Client;
+
+    @BeforeAll
```
massdosage, line 49: To match the `S3ProxyRuleTest` this should be `@Before`

massdosage (changes requested):

src/main/java/org/gaul/s3proxy/junit/S3ProxyExtension.java
```diff
@@ -0,0 +1,108 @@
...
+        core = new S3ProxyJunitCore(builder.builder);
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    @Override
+    public void beforeEach(ExtensionContext extensionContext) throws Exception {
+        core.beforeAll();
```
massdosage, line 89:
```suggestion
        core.before();
```
or `beforeEach();`

src/main/java/org/gaul/s3proxy/junit/S3ProxyExtension.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public void afterEach(ExtensionContext extensionContext) {
+        core.afterAll();
```
massdosage, line 94:
```suggestion
        core.after();
```
or `afterEach()`

pom.xml
```diff
@@ -360,6 +365,13 @@
       <!-- Required for S3ProxyRule -->
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <version>5.7.0</version>
+      <!-- Required for Junit5 Extension -->
```
massdosage, line 372:
```suggestion
      <!-- Required for S3ProxyExtension -->
```

src/main/java/org/gaul/s3proxy/junit/S3ProxyRule.java
```diff
@@ -45,60 +35,44 @@
     private static final Logger logger = LoggerFactory.getLogger(
```
massdosage, line 35: I don't think is used any more so you can remove it. From Maven:
```
[WARNING] /workspace/s3proxy/src/main/java/org/gaul/s3proxy/junit/S3ProxyRule.java:[35,33] [UnusedVariable] The field 'logger' is never read.
```

src/main/java/org/gaul/s3proxy/junit/S3ProxyJunitCore.java
```diff
@@ -0,0 +1,179 @@
...
+import org.jclouds.blobstore.BlobStore;
+import org.jclouds.blobstore.BlobStoreContext;
+import org.jclouds.blobstore.domain.StorageMetadata;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class S3ProxyJunitCore {
+
+    private static final Logger logger = LoggerFactory.getLogger(
+            S3ProxyExtension.class);
```
massdosage, line 39:
```suggestion
            S3ProxyJunitCore.class);
```

src/test/java/org/gaul/s3proxy/junit/S3ProxyExtensionTest.java
```diff
@@ -0,0 +1,93 @@
...
+            .builder()
+            .withCredentials("access", "secret")
+            .build();
+
+    private static final String MY_TEST_BUCKET = "my-test-bucket";
+
+    private static AmazonS3 s3Client;
+
+    @BeforeEach
+    public static void setUp() throws Exception {
```
massdosage, line 50: This test no longer runs as you can't apply `@BeforeEach` to a static method. Please run `mvn test` on the command line to make sure the whole build still passes.

src/test/java/org/gaul/s3proxy/junit/S3ProxyExtensionTest.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright 2014-2020 Andrew Gaul <andrew@gaul.org>
```
massdosage, line 2:
```suggestion
 * Copyright 2014-2021 Andrew Gaul <andrew@gaul.org>
```

src/main/java/org/gaul/s3proxy/junit/S3ProxyJunitCore.java
```diff
@@ -0,0 +1,179 @@
+/*
+ * Copyright 2014-2020 Andrew Gaul <andrew@gaul.org>
```
massdosage, line 2:
```suggestion
 * Copyright 2014-2021 Andrew Gaul <andrew@gaul.org>
```

src/main/java/org/gaul/s3proxy/junit/S3ProxyExtension.java
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright 2014-2020 Andrew Gaul <andrew@gaul.org>
```
massdosage, line 2:
```suggestion
 * Copyright 2014-2021 Andrew Gaul <andrew@gaul.org>
```

src/main/java/org/gaul/s3proxy/junit/S3ProxyJunitCore.java
```diff
@@ -0,0 +1,179 @@
...
+        private AuthenticationType authType = AuthenticationType.NONE;
+        private String accessKey;
+        private String secretKey;
+        private String secretStorePath;
+        private String secretStorePassword;
+        private int port = -1;
+        private boolean ignoreUnknownHeaders;
+        private String blobStoreProvider = "filesystem";
+
+        Builder() { }
```
massdosage, line 62: I think you can remove this constructor, the compiler will generate one, not sure we need a default one?

massdosage (changes requested):

gaul:
@xgourmandin could you address the review comments and rebase your PR?

xgourmandin:
> 
> 
> @xgourmandin could you address the review comments and rebase your PR?

Hi guys, 

sorry for te elay, I've been busy recently. I'll be able to implement this very soon.

gaul:
@xgourmandin Any chance you can finish this PR?  I would like to run a new release in a week or two.

xgourmandin:
Hi guys, 

Sorry for the absurdly long delay.
Only the Beta annotation is not settled yet.

gaul:
@massdosage Could you help review this?

gaul:
@xgourmandin Do you need to rebase this PR?

xgourmandin:
I have merged current develop branch so i don't think so.

pom.xml
```diff
@@ -367,6 +372,13 @@
       <!-- Required for S3ProxyRule -->
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <version>5.7.0</version>
```
massdosage, line 378:
```suggestion
      <version>5.8.1</version>
```
Might as well go with the latest release.

src/main/java/org/gaul/s3proxy/junit/S3ProxyJunitCore.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    S3ProxyJunitCore(Builder builder) {
+        accessKey = builder.accessKey;
+        secretKey = builder.secretKey;
+
+        Properties properties = new Properties();
+        try {
+            blobStoreLocation = Files.createTempDirectory("S3ProxyRule")
```
massdosage, line 107:
```suggestion
            blobStoreLocation = Files.createTempDirectory("S3Proxy")
```
(since this is no longer specific to just the rule)

src/test/java/org/gaul/s3proxy/junit/S3ProxyExtensionTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @RegisterExtension
+    static final S3ProxyExtension EXTENSION = S3ProxyExtension
+            .builder()
+            .withCredentials("access", "secret")
+            .build();
+
+    private static final String MY_TEST_BUCKET = "my-test-bucket";
+
+    private static AmazonS3 s3Client;
```
massdosage, line 47:
```suggestion
    private AmazonS3 s3Client;
```
for parity with S3ProxyRuleTest

src/test/java/org/gaul/s3proxy/junit/S3ProxyExtensionTest.java
```diff
@@ -0,0 +1,93 @@
...
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import com.amazonaws.services.s3.model.Bucket;
+import com.amazonaws.services.s3.model.S3ObjectSummary;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+
```
massdosage, line 36:
```suggestion
/**
 * This is an example of how one would use the S3Proxy JUnit extension in a unit
 * test as opposed to a proper test of the S3ProxyExtension class.
 */
```

src/main/java/org/gaul/s3proxy/junit/S3ProxyRule.java
```diff
@@ -16,89 +16,58 @@
...
-import org.jclouds.blobstore.domain.StorageMetadata;
 import org.junit.rules.ExternalResource;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * A JUnit Rule that manages an S3Proxy instance which tests can use as an S3
  * API endpoint.
  */
 @Beta
```
massdosage, line 30: @gaul - do we still want this marked as `Beta`? If so then I guess the Extension should have this annotation too?

massdosage (changes requested):
Looks mostly good, nice work, just a few small comments from me.

gaul:
Thank you for your contribution and patience @xgourmandin!  I plan to run a release this month if you want to include other enhancements.

PR #361, opened 2021-07-20
BlobStoreLocator Loader

kortemik:
Please do not merge, this is for comments on the BlobStoreLocator Loader. Is this something desired to make the blobstorelocator implementation modular or should the approach be different?

Please ignore the getter/setter changes, they are not relevant to this.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/AnonBlobStoreLocator.java
+++ b/src/main/java/org/gaul/s3proxy/AnonBlobStoreLocator.java
@@ -1,0 +1,22 @@
+package org.gaul.s3proxy;
+
+import org.jclouds.blobstore.BlobStore;
+
+import java.util.Map;
+
+
+public class AnonBlobStoreLocator implements AnonymousBlobStoreLocator {
+
+    Map.Entry<String, BlobStore> anonymousBlobStore;
+
+    @Override
+    public Map.Entry<String, BlobStore> locateBlobStore(
+            String identityArg, String container, String blob) {
+        return anonymousBlobStore;
+    }
+
+    @Override
+    public void setBlobStoreMap(Map.Entry<String, BlobStore> anonymousBlobStore) {
+        this.anonymousBlobStore = anonymousBlobStore;
+    }
+}
--- a/src/main/java/org/gaul/s3proxy/AnonymousBlobStoreLocator.java
+++ b/src/main/java/org/gaul/s3proxy/AnonymousBlobStoreLocator.java
@@ -1,0 +1,11 @@
+package org.gaul.s3proxy;
+
+import org.jclouds.blobstore.BlobStore;
+
+import java.util.Map;
+
+public interface AnonymousBlobStoreLocator extends BlobStoreLocator {
+
+    void setBlobStoreMap(Map.Entry<String, BlobStore> anonymousBlobStore);
+
+}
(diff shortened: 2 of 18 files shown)


gaul:
I don't understand what all this code is for but why can't you call `S3Proxy.setBlobStoreLocator` with your `BlobStoreLocator` implementation?

kortemik:
Please forgive me not attaching a brief explanation of the intention. I aim to create a feature which allows a user of the s3proxy to load their implementation of the BlobStoreLocator without modifying s3proxy code at all, meaning that they don't even have to compile the s3proxy themselves even if the s3proxy version changes as long as the API for the BlobStoreLocator is the same. This would be achived by using a ServiceLoader which can load the user implementation from the classpath and select the implementation by a system property. I completely understand that this kind of usage may be out of the goals of the project as many may be using s3proxy for testing purposes but we have aim to use it to create virtual ACLs on backends where only one credential is available for use augmented by a dedicated ACL database.

I used following information when digging into this:
```
https://docs.oracle.com/javase/tutorial/ext/basics/spi.html
https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html
https://www.baeldung.com/java-spi
https://www.logicbig.com/tutorials/core-java-tutorial/java-se-api/service-loader.html
https://stackoverflow.com/questions/17449924/service-loader-implementations-needing-default-public-constructors
https://www.tutorialspoint.com/java/util/serviceloader_iterator.htm
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html
https://sodocumentation.net/java/topic/5433/serviceloader
https://stackoverflow.com/questions/10304100/serviceloader-to-find-implementations-of-an-interface
https://stackoverflow.com/questions/20474206/choose-implementation-at-runtime
https://www.infoq.com/articles/java11-aware-service-module/
```

gaul:
I don't understand why anyone needs to recompile S3Proxy to use it as a library?  If the user just implements a `BlobStoreLocator` implementation and calls `S3Proxy.setBlobStoreLocator` it should just work.  Several users have done this in the past.  I don't understand why a service loader is required?

kortemik:
I mean not to use it as a library but as the main application with modular functionality. Would this be out of the question?

gaul:
The most common use case is running S3Proxy as the main application in a Docker container so I'm not opposed to the use case.  This PR is a mess so it is not clear what you are trying to do but if you just want anonymous authentication you just specify `s3proxy.authorization=none`.

kortemik:
Thank you for reviewing it. I decided to use the S3Proxy as a library though. I think the changes in this PR increase only complexity of the overall code and the users who want to implement their own rather find it easier to use the library interface instead of using a module interface as ClassLoader modules may be bit unknown approach although they may become more popular with java 11.

The aim was to be able to give out the possibility to just wget the s3proxy.jar and then
```
java -cp myBlobStoreLocator.jar:s3proxy.jar -Ds3proxy.blobstore.locator.impl_single=com.example.UserDefinedBlobStoreLocator org.gaul.s3proxy.Main
```

where user would have a project which creates a myBlobStoreLocator.jar with defined implementation of the SingleBlobStoreLocator  and declaring it as exported by having a file for the ClassLoader /META-INF/services/org.gaul.s3proxy.SingleBlobStoreLocator with line
```
com.example.UserDefinedBlobStoreLocator
```

May be it will be handy for later study but I think it is too much for now.

gaul:
A few years ago I looked into how to extend S3Proxy with arbitrary JavaScript so that users could change behavior without recompilation:

https://github.com/gaul/s3proxy/tree/javascript

My concern was that it was actually easier to just extend S3Proxy as a library since otherwise you needed to understand the Java _and_ JavaScript parts.

Issue #364, opened 2021-08-17
Support different ak/sk for per Bucket?


whybeyoung:
After read the doc, i don't find that how to set different ak/sk for per bucket. if not support, please suggest me how to adapt  to this case.  I'm not prefer start multiple s3proxy process to deal with it, it's very expensive

gaul:
@timuralp I think you are looking into this?

berlinsaint:
My case is a mulit teanant case, and i want to provide user with different bucket, every user can control their own bucket.

Now the s3proxy only provide one pair ak/sk, which has data leak in our case...

I thinks it's a useful feature in s3proxy...

timuralp:
I'm investigating using the bucket to look up the blobstore in the blobstore locator. The s3proxy configuration would look like a collection of properties files, one per tenant, with a set of buckets per blobstore. @berlinsaint is that similar to your ask?

berlinsaint:
> I'm investigating using the bucket to look up the blobstore in the blobstore locator. The s3proxy configuration would look like a collection of properties files, one per tenant, with a set of buckets per blobstore. @berlinsaint is that similar to your ask?

It's similar, but not  the same.  if use multiple properties you said（and i have saw that indeed）, means it will restart or then reload the config properties , which sounds like not very graceful. in additionaly, it will provide a set of buckets not only one ...

I saw that the doc mentioned that not support acl, disappointed a little... maybe this feature is related to “multiple user and give them differenct bucket access"?

Jayd603:
I get what whybeyoung is trying to do here, he wants s3proxy to handle many different properties files (and thus credentials that go to different buckets/backends) and be able to add/remove them gracefully without a hard restart of s3proxy that disrupts connections.  gaul - how difficult do you think that would be?

gaul:
This works today, and the machinery landed two weeks after this was filed:
@timuralp's bucket locator (558efb7, released in 1.9.0) lets one process hold
several configurations, each with its own front-end key pair, its own backend
and its own set of buckets. I never closed the issue.

One properties file per tenant:

```properties
# tenant-a.conf
s3proxy.endpoint=http://127.0.0.1:8080
s3proxy.authorization=aws-v2-or-v4
s3proxy.identity=tenant-a-key
s3proxy.credential=tenant-a-secret
s3proxy.bucket-locator.1=bucket-a
jclouds.provider=filesystem
jclouds.filesystem.basedir=/srv/tenant-a
```

```properties
[...]

PR #365, opened 2021-08-19
Add a experimental feature that can get use different ak\sk for per bucket.

berlinsaint:
As issue  [`#364`](https://github.com/gaul/s3proxy/issues/364) says, i add some code implement to achieve my request.

Please take a look, It's no need to be merged.  But you can consider  it, if you permit, i can do it better（but not this pr）

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -457,5 +457,12 @@
       <artifactId>sqlite-jdbc</artifactId>
       <version>3.36.0.1</version>
     </dependency>
+    <!-- https://mvnrepository.com/artifact/org.yaml/snakeyaml -->
+    <dependency>
+      <groupId>org.yaml</groupId>
+      <artifactId>snakeyaml</artifactId>
+      <version>1.29</version>
+    </dependency>
+
   </dependencies>
 </project>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -452,5 +452,10 @@
       <artifactId>javax.annotation-api</artifactId>
       <version>1.3.2</version>
     </dependency>
+    <dependency>
+      <groupId>org.xerial</groupId>
+      <artifactId>sqlite-jdbc</artifactId>
```
gaul, line 457: S3Proxy cannot take on a [native dependency](https://github.com/xerial/sqlite-jdbc#how-does-sqlitejdbc-work) which cannot be used in some environments.

Instead of storing configuration in a database, why not use a properties or JSON file?  You can send a `SIGHUP` to the S3Proxy process to force it to re-read the file, similar to other Unix daemons.

berlinsaint: In my case, i need a programmable  way  to control bucket access' add/delete,  and we need to have HA feature, so i use db and cache.
BTW, for your suggestion, i can add an YAML/JSON cod to implete  Interface: AccessSecretManager. how  do you think?

berlinsaint:
will submit another pr some days. closed now

PR #366, opened 2021-08-20
Support Iso8601 in date headers

peterzeller:
Background: I was trying to use the [S3 backup feature of FoundationDB](https://apple.github.io/foundationdb/backups.html) with s3proxy to do backups on Azure. This did not work, because the date header sent by the backup tool is in Iso8601 format.

This PR adds code to parse the date header in Iso8601 if the default header parsing fails.


gaul:
Thank you for your contribution @peterzeller!  If you have time, can you see why [s3-tests](https://github.com/ceph/s3-tests) didn't find this earlier and file an issue there?

Issue #368, opened 2021-08-23
request and response log


kortemik:
I am willing to make a PR for request and response log feature.

Would it be preferred that one adds:

1) A separate handler in addition to S3ProxyHandlerJetty like Jetty describes for NCSA at https://www.eclipse.org/jetty/documentation/jetty-9/index.html#configuring-jetty-logging 

2) Modifies the S3ProxyHandlerJetty so that it takes a logging class implementatation as argument and modify S3Proxy so it has setLogger() which passes the logging class implementatation to S3ProxyHandlerJetty? The default would be a no-op final class which gets optimized out.

gaul:
One simple implementation would to be to add `logger.trace` calls for the request and response codes and headers.  What value does a logger class provide?

Duplicate of #213?

kortemik:
Logger class provides the ability to log needed headers and such, of course we could change that the logger.trace outputs them all as json but it may be too much for request log.

PR #369, opened 2021-08-23
AccessLogger

kortemik:
#368

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyAccessLogger.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyAccessLogger.java
@@ -1,0 +1,17 @@
+package org.gaul.s3proxy;
+
+import org.eclipse.jetty.server.Request;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public final class S3ProxyAccessLogger implements AccessLogger {
+    @Override
+    public void logRequest(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) {
+        return;
+    }
+
+    @Override
+    public void logResponse(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/gaul/s3proxy/S3ProxyHandlerJetty.java
```diff
@@ -153,6 +158,7 @@ public void handle(String target, Request baseRequest,
                 throw throwable;
             }
         }
+        accessLogger.logResponse(target, baseRequest, request, response);
```
gaul, line 161: This doesn't execute on any exception paths because there are several `return` statements.

src/main/java/org/gaul/s3proxy/S3ProxyAccessLogger.java
```diff
@@ -0,0 +1,17 @@
+package org.gaul.s3proxy;
+
+import org.eclipse.jetty.server.Request;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public final class S3ProxyAccessLogger implements AccessLogger {
+    @Override
+    public void logRequest(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) {
+        return;
```
gaul, line 10: This doesn't seem to _do_ anything.  I would expect it to at least include the logging in `S3ProxyHandler.doHandle`:

```
logger.debug("request: {}", request);
logger.debug("header: {}: {}", headerName, Strings.nullToEmpty(headerValue));
```

gaul:
@kortemik Any updates on this?  I hope to run a release this month.

kortemik:
@gaul I am afraid that the timeline is very tight for me, on which date would it be? If I fix the proposed things I could manage but if more changes are requested, it will be bit hard.

gaul:
Please reopen if you find a path forward here.

Issue #371, opened 2021-08-27
Use container names to disambiguate backend stores


timuralp:
s3proxy disambiguates backend stores by different credentials, which means that from the application side typically a new client needs to be created to connect to the different backends. To reduce friction, it would be great to be able to configure s3proxy with multiple backends with the same credentials and specify the buckets associated with them.

The specific use case is Azure, where multiple storage accounts are common, but being able to hide that complexity from the application would simplify access.

PR #376, opened 2021-09-12
Allow differentiating blobstores by buckets

timuralp:
Allows differentiating the backend blobstores by configured buckets.
The commit allows using the same s3proxy credentials across the
different backends and relies on the bucket names to disambiguate them.

The "default" blobstore is the one that occurs first.

Fixes: #371


src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -117,14 +121,44 @@ public static void main(String[] args) throws Exception {
...
+                localIdentity = properties.getProperty(
                         S3ProxyConstants.PROPERTY_IDENTITY);
-                String localCredential = properties.getProperty(
+                localCredential = properties.getProperty(
                         S3ProxyConstants.PROPERTY_CREDENTIAL);
-                locators.put(localIdentity, Maps.immutableEntry(
-                        localCredential, blobStore));
+                Map.Entry<String, String> key = Maps.immutableEntry(
+                        localIdentity, null);
+                if (!parsedLocations.contains(key)) {
```
gaul, line 142: You can use the return value from `parsedLocations.add`.

gaul (changes requested):

gaul:
@berlinsaint @jiffysrc there seems to be some overlap between #372, #374, and this PR.  Can you guys sort out which approach is the best?

src/main/java/org/gaul/s3proxy/Main.java
timuralp: Done

src/main/java/org/gaul/s3proxy/S3ProxyConstants.java
```diff
@@ -66,6 +66,9 @@
             "s3proxy.max-single-part-object-size";
     public static final String PROPERTY_V4_MAX_NON_CHUNKED_REQUEST_SIZE =
             "s3proxy.v4-max-non-chunked-request-size";
+    /** Used to locate blobstores by specified bucket names. */
+    public static final String PROPERTY_BUCKET_LOCATOR =
+            "s3proxy.bucket-locator";
```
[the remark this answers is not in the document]
timuralp: Expanded the description. The bucket-locator property specifies the bucket names associated with the backend specified in the properties file. Different properties files can specify different buckets, allowing a client to use the same S3Proxy credentials, but have the requests route to different backends.

timuralp:
I think #374 and this PR are related, but orthogonal: adding external authorization callback vs being able to select a backend by the bucket name. I'm struggling to understand the intent of #372, aside from introducing alternate stores for credentials.

berlinsaint:
> I think #374 and this PR are related, but orthogonal: adding external authorization callback vs being able to select a backend by the bucket name. I'm struggling to understand the intent of #372, aside from introducing alternate stores for credentials.

agree

src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -117,14 +121,42 @@ public static void main(String[] args) throws Exception {
...
-                locators.put(localIdentity, Maps.immutableEntry(
-                        localCredential, blobStore));
+                Map.Entry<String, String> key = Maps.immutableEntry(
+                        localIdentity, null);
+                if (parsedLocations.add(key)) {
+                    locators.put(key, Maps.immutableEntry(
+                            localCredential, blobStore));
+                }
+            }
+            if (!buckets.isEmpty()) {
```
gaul, line 147: Do you need to check whether buckets is empty?  The foreach loop should handle this already.

gaul (approved):

gaul:
Tested this with two local filesystem configs and verified that objects go where they should.  Also tried some duplicate buckets and the error message is comprehensible.

gaul:
Superseded by #378.

PR #378, opened 2021-10-19
Allow differentiating blob stores by buckets

timuralp:
Allows differentiating the backend blob stores by configured buckets.
The commit allows using the same s3proxy credentials across the
different backends and relies on the bucket names to disambiguate them.

If no buckets are configured, the prior blob store selection algorithm
is used, which returns the first configured blob store for the specified
identity.

The patch supports the glob syntax, which allows specifying groups of
buckets more easily. However, there is no checking for overlapping
globs.

Fixes: #371


timuralp:
This PR conflicts with #376 and only one of these should be merged. The difference is that this PR implements glob support, whereas #376 only matched on exact bucket names.

PR #383, opened 2021-10-29
Publish all platforms that the base image supports

### LINKED ISSUE #382: Support arm64
The latest MacBook uses the M1 CPU which is Apple Silicon running arm64. Currently this image only supports `linux/amd64`. It should be fairly easy to support multiple architectures: https://docs.docker.com/desktop/multi-arch/

ansman:
This fixes #382

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ script:
  # Work around Travis environment failure: https://github.com/travis-ci/travis-ci/issues/7940
  - export BOTO_CONFIG=/dev/null
  - ./src/test/resources/run-s3-tests.sh
- - docker build -f Dockerfile -t $REPO:$COMMIT .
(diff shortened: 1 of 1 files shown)


gaul:
@ansman Please investigate CI failures -- maybe a typo?

ansman:
I'm guessing `buildx` isn't available on the Travis docker. I'll see if it can be installed.

ansman:
So apparently it's quite the pain to get this to work with Travis: https://www.docker.com/blog/multi-arch-build-what-about-travis/

Would you be open to switching to GitHub Actions instead of Travis? I can provide a PR for that as well if you'd like.

gaul:
I am not wild about the changes Travis has made over the last few years and am mostly an inertia/legacy user.  I use GitHub actions on another project so I am not opposed to switching.  If you can provide a GitHub action with parity with the existing config then I don't mind switching.

usmanzaheer1995:
Hi! I was wondering if there is any news on this? I am trying to run the s3proxy with docker on my M1 machine (platform `darwin/arm64`)

gaul:
I believe this is superseded by #391.  @ansman can you confirm and close this if so?

PR #384, opened 2021-11-04
Replace the travis scripts with GitHub Actions

ansman:


ansman:
@gaul This is ready for review now. You can see a successful run here: https://github.com/ansman/s3proxy/runs/4110086097

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,24 @@
...
+          submodules: 'recursive'
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: '11'
+      - name: Install dependencies
+        run: sudo apt-get install -y libevent-dev python3-pip python3-virtualenv
+      - name: Run Tests Tests
+        run: |
+          mvn package
```
ansman, line 18: I'm not sure why I have to run `mvn package` while the travis script only ran `mvn test` but it seems that `s3proxy` isn't generated when I run `mvn package`

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,24 @@
...
+          ./src/test/resources/run-s3-tests.sh
+      - name: Build Docker Image
+        run: ./publish_img.sh
+        env:
+          DOCKER_USER: ${{ secrets.DOCKER_USER }}
+          DOCKER_PASS: ${{ secrets.DOCKER_PASS }}
```
ansman, line 24: Make sure to add these two secrets before merging

.github/workflows/run-tests.yml
gaul: I suspect that this is just some difference between Travis and Docker.  We might want to use `mvn verify` but let's follow up in a subsequent PR.

.github/workflows/run-tests.yml
gaul: I added secrets.  Let's merge this then try to force a build.

gaul: Could you take a look at https://github.com/gaul/s3proxy/runs/4211919696?check_suite_focus=true#step:6:18402 ?  I suspect that `REPO` needs to be set.

publish_img.sh
```diff
@@ -1,7 +1,12 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
gaul, line 1: Might want `set -o unset` here.

PR #386, opened 2021-11-05
Add required Content-MD5 middleware

gaul:
This enforces that the Content-MD5 header is present in PutObject and
UploadPart requests.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/Main.java
+++ b/src/main/java/org/gaul/s3proxy/Main.java
@@ -240,6 +240,13 @@ public final class Main {
             blobStore = ReadOnlyBlobStore.newReadOnlyBlobStore(blobStore);
         }
 
+        String requiredContentMD5BlobStore = properties.getProperty(
+                S3ProxyConstants.PROPERTY_REQUIRED_CONTENT_MD5_BLOBSTORE);
+        if ("true".equalsIgnoreCase(requiredContentMD5BlobStore)) {
+            System.err.println("Using required Content-MD5 storage backend");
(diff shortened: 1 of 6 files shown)


src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -240,6 +240,13 @@ private static BlobStore parseMiddlewareProperties(BlobStore blobStore,
             blobStore = ReadOnlyBlobStore.newReadOnlyBlobStore(blobStore);
         }
 
+        String requiredContentMD5BlobStore = properties.getProperty(
+                S3ProxyConstants.PROPERTY_READ_ONLY_BLOBSTORE);
```
timuralp, line 244: Should this be `PROPERTY_REQUIRED_CONTENT_MD5_BLOBSTORE`?

gaul: Done.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2654,6 +2657,7 @@ private void handleCopyPart(HttpServletRequest request,
                         his.hash().asBytes());
             } else {
                 Payload payload = Payloads.newInputStreamPayload(is);
+                // TODO: Content-MD5
```
gaul, line 2660: Running s3fs tests I found that this middlware cannot handle UploadPartCopy due to the emulation that S3Proxy does.  It cannot get the correct MD5 of a range of data to satisfy the underlying `RequiredContentMD5BlobStore`.

gaul:
Closing as infeasible.

PR #388, opened 2021-11-15
Fix a missing variable when publishing

ansman:


gaul:
Follows on to #384.

publish_img.sh
```diff
@@ -4,9 +4,9 @@ docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
 docker buildx build --platform linux/amd64 -t $GITHUB_REPOSITORY:latest .
 
 if [[ "$GITHUB_EVENT_NAME" == "push" && $GITHUB_REF == refs/heads/master ]]; then
-  docker tag $REPO:latest $GITHUB_REPOSITORY:${GITHUB_SHA::8}
+  docker tag $GITHUB_REPOSITORY:latest $GITHUB_REPOSITORY:${GITHUB_SHA::8}
   docker push --all-tags $GITHUB_REPOSITORY
 elif [[ "$GITHUB_EVENT_NAME" == "create" && $GITHUB_REF == refs/tags/* ]]; then
-  docker tag $REPO:latest $GITHUB_REPOSITORY:${GITHUB_REF#refs/tags/}
+  docker tag $GITHUB_REPOSITORY:latest $GITHUB_REPOSITORY:${GITHUB_REF#refs/tags/}
```
gaul, line 10: I pushed a commit to change this to `REPO` which refers to Docker Hub andrewgaul/s3proxy instead of GitHub gaul/s3proxy.  But this is still broken for some reason:

https://github.com/gaul/s3proxy/runs/4218962196?check_suite_focus=true#step:6:17852

I set `DOCKER_USER` and `DOCKER_PASS` set in the repo's secrets.  Any other ideas?

ansman: I saw that, it's very strange as it should work.

Can you post a picture of the GH Actions secrets page (the credentials should be blanked out so it should be safe to post)

ansman: Ah, I think it might be logging in to the wrong "area". I opened #390, see if that helps

ansman: Still broken. I have no idea why. Does it work if you deploy locally (setting the `DOCKER_USER`, `DOCKER_PASS`, `GITHUB_EVENT_NAME`,  `GITHUB_REF` and  `GITHUB_SHA` manually)?

ansman: I opened a PR where I migrate to the official docker steps instead: https://github.com/gaul/s3proxy/pull/391

ansman: Ok, still failing. Not sure why. I found this issue: https://github.com/docker/login-action/issues/106

Do you have MFA enabled and are you using the access token as the password?

gaul: This now succeeds -- I had previously configured this as an environment secret but apparently the correct solution is a repository secret.  Please test the ARM image and thank you for your contribution!

ansman: Ah, good catch!

PR #391, opened 2021-11-16
Start publishing arm64

### LINKED ISSUE #382: Support arm64

ansman:
Also switch to the official docker steps in GitHub Actions.

This fixes #382


.github/workflows/run-tests.yml
```diff
@@ -17,8 +29,26 @@ jobs:
...
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASS }}
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: andrewgaul/s3proxy
+          tags: |
+            type=sha,event=branch
```
ansman, line 44: This tags the image with the short SHA when pushing to `master`

.github/workflows/run-tests.yml
```diff
@@ -17,8 +29,26 @@ jobs:
...
+            type=semver,pattern={{version}}
```
ansman, line 45: This tags the image with the full version when pushing a git tag.

It also tags it with `latest` when pushing a git tag

.github/workflows/run-tests.yml
```diff
@@ -17,8 +29,26 @@ jobs:
...
+            type=semver,pattern={{major}}.{{minor}}
```
ansman, line 46: This tags it with the `major.minor` version when pushing a git tag

.github/workflows/run-tests.yml
```diff
@@ -17,8 +29,26 @@ jobs:
...
+            type=semver,pattern={{major}}
```
ansman, line 47: This tags it with the `major` version when pushing a git tag

PR #393, opened 2021-11-26
Update build

cstamas:
In general update the ancient build plugins
around.

Changes:
* bring all maven plugins to latest versions
* update checkstyle (as pre 8.41 is not supported anymore)
* update checkstyle rules
* make checkstyle plugin not fail the build (as there are issues)
* fix maven-shade-plugin warnings about dupe resource
* fix modernizer issue as it fails the build
* make build reproducible, see https://maven.apache.org/guides/mini/guide-reproducible-builds.html

Added comment about parent POM: the currently used parent is
10 years old, moreover, if you look in Maven Central,
the GA is unmaintained (last release was in 2014). Better introduce
own parent POM, or just copy whatever needed into this POM
(related to staging) and stop using the ancient POM.


gaul:
I verified the PR locally since CI did not run.  Before S3Proxy 1.8.0, the project maintained compatibility with Java 7 and thus had a lot of outdated dependencies.  Thank you for paying down this technical debt @cstamas!

Issue #126, opened 2016-02-24
Add mechanism to encrypt object data before sending to storage backend


andrewgaul:
This would enable data privacy on backends without encryption, e.g., Rackspace Cloud Files, as well as improve privacy on backends with it, e.g., Amazon S3, due to handling the private key in S3Proxy.

shenghu:
@andrewgaul does s3proxy support encryption on object storage?

andrewgaul:
@shenghu S3Proxy does not support object server-side encryption due to a lack of support in the underlying jclouds, tracked by [JCLOUDS-1253](https://issues.apache.org/jira/browse/JCLOUDS-1253).  This issue tracks something different, object client-side encryption.  This difference lies in who does key management; in the former, the object store manages keys, in the latter, S3Proxy manages keys.  Ideally S3Proxy would support both.

andrewgaul:
S3Proxy could include `EncryptedBlobStore` from Bounce:

https://github.com/bouncestorage/bouncestorage/blob/master/bounce/src/main/java/com/bouncestorage/bounce/EncryptedBlobStore.java

PR #398, opened 2022-01-08
Support for transparent encryption

FlorinPeter:
First of all I would like to thank @gaul for this nice project!

This PR will introduce the support for transparent encryption.
More information can be found in ```docs/Encryption.md```

I would love to get some feedback and hopefully see this PR merged ;)


gaul:
Can this be built as a middleware, similar to #296?  References #126.

FlorinPeter:
Thx for the hint @gaul!
I will have a look.

FlorinPeter:
@gaul sorry that I missed the middleware part.
Updated the branch and now it is implemented as ```EncryptedBlobStore``` middleware.

FlorinPeter:
I improved the code coverage and added the missing salt for the key generation.

FlorinPeter:
@gaul I found some regressions and need to look into it this is why I set the PR back to draft

FlorinPeter:
@gaul I hope that the implementation now is way better after rework on some internal parts.

FlorinPeter:
I added many fixes after testing with a docker registry that uses s3proxy to encrypt the blobs:
- fixed offset calculation 
- removed debug log 
- changed the SslContextFactory to SslContextFactory.Server
- handleListMultipartUploads now supports prefix
- added EncryptedBlobStoreLiveTest

FlorinPeter:
Added my last fixes from various findings:
- azure and gcp multipart list return now reliable upload ids
- removed previous decryption pipes by ```DecryptionInputStream``` and speed up decryption by factor 10
- fix handleCopyPart for gcp

FlorinPeter:
@gaul can you have a look?

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,784 @@
...
+    private SecretKeySpec secretKey;
+
+    private EncryptedBlobStore(BlobStore blobStore, Properties properties) {
+        super(blobStore);
+
+        String password = properties.getProperty(
+            S3ProxyConstants.PROPERTY_ENCRYPTED_BLOBSTORE_PASSWORD);
+        String salt = properties.getProperty(
+            S3ProxyConstants.PROPERTY_ENCRYPTED_BLOBSTORE_SALT);
+        initStore(password, salt);
```
timuralp, line 75: I'd add `checkNotNull` here, as the NPE resulting from forgetting to set the salt or password is not as obvious to parse, e.g.:
```
Using encrypted storage backend
Exception in thread "main" java.lang.RuntimeException: java.lang.NullPointerException
	at org.gaul.s3proxy.EncryptedBlobStore.initStore(EncryptedBlobStore.java:93)
	at org.gaul.s3proxy.EncryptedBlobStore.<init>(EncryptedBlobStore.java:75)
	at org.gaul.s3proxy.EncryptedBlobStore.newEncryptedBlobStore(EncryptedBlobStore.java:80)
	at org.gaul.s3proxy.Main.parseMiddlewareProperties(Main.java:264)
	at org.gaul.s3proxy.Main.main(Main.java:123)
Caused by: java.lang.NullPointerException
	at org.gaul.s3proxy.EncryptedBlobStore.initStore(EncryptedBlobStore.java:88)
	... 4 more
```

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,784 @@
...
+            return decryptBlob(decryption, containerName,
+                delegate().getBlob(containerName, blobName, getOptions));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public String putBlob(String containerName, Blob blob) {
+        return delegate().putBlob(containerName,
```
timuralp, line 450: It would be nice to tee the input stream through the MD5 hash, so that the returned ETag would match the content rather than the encrypted content. Otherwise common clients like s3cmd will break, e.g.:
```
DEBUG: signature-v4 headers: {'content-length': '6943', 'content-type': 'text/plain', 'x-amz-meta-s3cmd-attrs': 'atime:1642891263/ctime:1635210219/gid:1000/gname:timur/md5:d851162cc8f800d496c800efee338e89/mode:33188/mtime:1635210219/uid:1000/uname:timur', 'x-amz-storage-class': 'STANDARD', 'x-amz-date': '20220122T230420Z', 'Authorization': 'AWS4-HMAC-SHA256 Credential=test/20220122/us-east-1/s3/aws4_request,SignedHeaders=content-length;content-type;host;x-amz-content-sha256;x-amz-date;x-amz-meta-s3cmd-attrs;x-amz-storage-class,Signature=84b53372268cbdd61ab664d1f0e6c0b73aff05950d1cc517efba241b1082ef6c', 'x-amz-content-sha256': '08ad980cc2bf1b9ff3ff329ca6402eeb2a0131d17b590169e987b8dbd1060366'}
DEBUG: get_hostname(timur-s3): localhost:10080
DEBUG: ConnMan.get(): re-using connection: http://localhost:10080#1
DEBUG: format_uri(): /timur-s3/README.md
 6943 of 6943   100% in    0s     3.27 MB/sDEBUG: ConnMan.put(): connection put back to pool (http://localhost:10080#2)
DEBUG: Response:
{'data': b'',
 'headers': {'content-length': '0',
             'date': 'Sat, 22 Jan 2022 23:04:20 GMT',
             'etag': '"5405364c01976e87a31e476624d0dfeb"',
             'server': 'Jetty(9.4.41.v20210516)',
             'x-amz-request-id': '4442587FB7D0A2F9'},
 'reason': 'OK',
 'size': 6943,
 'status': 200}
 6943 of 6943   100% in    0s    25.79 KB/s  done
DEBUG: MD5 sums: computed=d851162cc8f800d496c800efee338e89, received=5405364c01976e87a31e476624d0dfeb
WARNING: MD5 Sums don't match!
```

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,784 @@
...
+        return delegate().getContext().unwrap().getProviderMetadata().getId();
+    }
+
+    private String generateUploadId(String container, String blobName) {
+        String path = container + "/" + blobName;
+        return DigestUtils.sha256Hex(path);
+    }
+
+    @Override
+    public Blob getBlob(String containerName, String blobName) {
```
timuralp, line 340: Can this call `getBlob(containerName, blobName, null)`?

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,784 @@
...
+                    Exception e =
+                        new Exception("multiple ranges are not supported");
+                    throw new RuntimeException(e);
+                }
+
+                String range = getOptions.getRanges().get(0);
+                String[] ranges = range.split("-", 2);
+                if (ranges[0].isEmpty()) {
+                    // clear current ranges to avoid multiple ranges
+                    getOptions.getRanges().clear();
```
timuralp, line 379: Hoist this above the `if` statement -- we have to update the range header regardless.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,784 @@
...
+
+                    // handle to read from the end a specific amount of bytes
+                    long length = Long.parseLong(ranges[1]);
```
timuralp, line 382: Consider assigning `offset` and `end` in the conditional clauses and move the common code (lines 383-393) outside of the conditional. We should be able to use `getOptions.range()` in all cases, right?

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,784 @@
...
+        try {
+            // adjust name if a blob with the .s3enc suffix exists
+            blobName = blobNameWithSuffix(containerName, blobName);
+
+            // init the decryption
+            Decryption decryption = null;
+            if (getOptions != null && getOptions.getRanges().size() > 0) {
+
+                // we support only one range
+                if (getOptions.getRanges().size() > 1) {
```
timuralp, line 369: This should never happen given the code here: https://github.com/gaul/s3proxy/blob/master/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java#L1712-L1714. I don't mind being defensive, but since S3 doesn't allow multiple ranges, I'd expect the situation to never happen.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,784 @@
...
+                        long endAt = decryption.calculateEndAt(end);
+                        getOptions.range(startAt, endAt);
+                    }
+                }
+            }
+
+            // initialize the decryption if no ranges were requested
+            if (decryption == null) {
+                decryption =
+                    new Decryption(secretKey, delegate(), containerName,
```
timuralp, line 437: Can we sink the instantiation of `Decryption` into `decryptBlob` and update its signature to:
`private Blob decryptBlob(String container, Blob blob)`? I think the `range` calculation is the only difference and maybe it could actually be done in the `Decryption` constructor itself (via passing the `Blob`, rather than just the name)?

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,784 @@
...
+    }
+
+    private boolean multipartRequiresStub() {
+        String blobStoreType = getBlobStoreType();
+        return Quirks.MULTIPART_REQUIRES_STUB.contains(blobStoreType);
+    }
+
+    private String blobNameWithSuffix(String container, String name) {
+        String nameWithSuffix = blobNameWithSuffix(name);
+        if (delegate().blobExists(container, nameWithSuffix)) {
```
timuralp, line 320: It would be better to avoid the HEAD request here. Can we always do a GET with the suffix and on 404 GET without it? I haven't internalized what changes would be required, but turning every GET into HEAD + GET is not desirable.

timuralp (changes requested):
This is a significant chunk work -- thank you for doing it! It takes a little bit longer to review, though.

I tested simple object operations, which work well. I'll look through the multipart handling in a bit and the encryption implementation itself in a little bit.

FlorinPeter:
@timuralp thx for looking into it, I will push my changes and squash after we are done with the review.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
FlorinPeter: since I tried to make it work in a mixed scenario where you can have encrypted and unencrypted files on the backend this is the trade of right now.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
FlorinPeter: I was also thinking about this but we need to adjust ```getOptions``` before it get passed to the delegate to get the blob.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
timuralp: Right, I understand the reasoning. I'm suggesting making an unconditional request for `<blob>.s3enc` and if it fails, making another request for `<blob>`, which in the case of encrypted blobs would result in a single request and in 2 requests for unencrypted blobs. Is it challenging to handle the error?

FlorinPeter: Got your point, will look tomorrow if I can find a way doing it.

FlorinPeter: I was reviewing the part and was only able to reduce one unnecessary HEAD call. 
As of now I don't see a possibility to get rid of the additional HEAD call therefore I extended the limitations in Encryption.md.
I hope that it is not a blocker from your side?

timuralp: I didn't notice the one more HEAD request -- thanks for fixing that up. I think we can table this for now and maybe fix it up afterward, as I have a few other points I wanted to discuss.

timuralp:
@FlorinPeter I was looking at the `crypto/Encryption` implementation and don't understand why we need to use a thread there. I expected it to wrap the supplied input stream and implements the `read()` method to produce the encrypted blocks (like the `CipherInputStream`) and padding. Can you expand on why we need the threads?

docs/Encryption.md
```diff
@@ -0,0 +1,76 @@
+S3Proxy
+
+# Encryption
```
gaul, line 3: Are there any other precedents that we can follow for S3 client-side encryption, e.g., from AWS SDK or some other project?

docs/Encryption.md
```diff
@@ -0,0 +1,76 @@
+S3Proxy
+
+# Encryption 
+
+## Motivation 
+The motivation behind this implementation is to provide a fully transparent and secure encryption to the s3 client while having the ability to write into different clouds.
+
+## Cipher mode
+The chosen cipher is ```AES/CFB/NoPadding``` because it provides the ability to read from an offset like in the middle of a ```Blob```.
+While reading from an offset the decryption process needs to consider the previous 16 bytes of the AES block.
```
gaul, line 10: Store the cipher type in user metadata?  This allows having mixed encryption types in case the data format ever changes.

docs/Encryption.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Multipart handling 
+A single ```Blob``` can be uploaded by the client into multiple parts. After the completion all parts are concatenated into a single ```Blob```.
+This procedure will result in multiple parts and paddings being held by a single ```Blob```.
+
+### Single blob example
+```
+-------------------------------------
+| ENCRYPTED BYTES         | PADDING |
+-------------------------------------
```
gaul, line 35: Is there padding?  Above it says `AES/CFB/NoPadding`.  Also what about the header you prepend (that I recommended removing)?

src/test/java/org/gaul/s3proxy/EncryptedBlobStoreLiveTest.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    @Override
+    public void testUpdateObjectCannedACL() {
+        throw new SkipException("blob access control not supported");
+    }
+
+    @Override
+    public void testCopyIfMatch() {
+        throw new SkipException("copy if match not supported");
```
gaul, line 285: If you get rid of the ETag munging will this test case pass?

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
gaul: Issuing HeadObject followed by GetObject opens the door to more eventual consistency with some object stores.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,794 @@
...
+    private SecretKeySpec secretKey;
+
+    private EncryptedBlobStore(BlobStore blobStore, Properties properties)
+        throws IOException {
+        super(blobStore);
+
+        String password = properties.getProperty(
+            S3ProxyConstants.PROPERTY_ENCRYPTED_BLOBSTORE_PASSWORD);
+        if (password == null || password.isEmpty()) {
+            throw new IOException(
```
gaul, line 78: Should this be an `IllegalArgumentException`?

gaul (changes requested):
This is a large PR so I haven't read all the code and tests.  I am initially concerned about the data format and whether you can use user metadata instead of blob suffixes.  Would preserve the original object size or if there would be padding?

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,794 @@
...
+
+    @Override
+    public String putBlob(String containerName, Blob blob,
+        PutOptions putOptions) {
+        // pass the stream through a md5 hashing to have a proper eTag
+        HashingInputStream his = blobHashing(blob);
+        delegate().putBlob(containerName,
+            encryptBlob(containerName, blob), putOptions);
+
+        return his != null ? his.hash().toString() : null;
```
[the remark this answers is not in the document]
FlorinPeter: yes this is true. I changed that because of Timurs finding now I reverted it to my first implementation.

src/main/java/org/gaul/s3proxy/crypto/Encryption.java
```diff
@@ -0,0 +1,97 @@
...
+        cipher.init(Cipher.ENCRYPT_MODE, key, iv);
+        cis = new CipherInputStream(isRaw, cipher);
+        part = partNumber;
+    }
+
+    @Override
+    public final void run() {
+        try {
+            long size = IOUtils.copyLarge(cis, pipeOut);
+            pipeOut.write(padding(iv, part, size));
```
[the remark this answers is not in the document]
FlorinPeter: Also Timur was pointing to that ;) I was also not happy with the thread approach and changed it and hope that it fits for you.

docs/Encryption.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Single blob decryption 
+First the ```BlobMetadata``` is requested to get the encrypted ```Blob``` size. The last 64 bytes of ```PartPedding``` are fetched and inspected to detect if a decryption is necessary.
+The cipher is than initialized with the IV and the key.
+
+### Multipart blob decryption 
+The process is similar to the single ```Blob``` decryption but with the difference that a list of parts is computed by fetching all ```PartPedding``` from end to the beginning.
+
+## Blob suffix
+Each stored ```Blob``` will get a suffix named ```.s3enc``` this helps to determine if a ```Blob``` is encrypted. For the s3 client the ```.s3enc``` suffix is not visible and the ```Blob``` size will always show the unencrypted size.
```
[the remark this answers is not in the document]
FlorinPeter: > @FlorinPeter I was looking at the `crypto/Encryption` implementation and don't understand why we need to use a thread there. I expected it to wrap the supplied input stream and implements the `read()` method to produce the encrypted blocks (like the `CipherInputStream`) and padding. Can you expand on why we need the thread?
> 
I was also not happy with that and changed it now.

> Also, have you considered using the object metadata vs object suffix for the encrypted blobs? I'm curious why you settled on the suffix approach -- the object metadata seems more elegant (no changes in blob names). Do you see difficulty from switching to using the metadata?

Yes I considered a lot of possibilities and this is now my 3rd or 4th iteration. 
The problem with the metadata is that on multipart uploads you need to know the metadata like size of all parts before initializing the upload this fact makes impossible to change the metadata of the concated blob with the final size and parts. I also played with content type and other things.
Right now I use the s3proxy for the docker registry to make sure that all layers are encrypted on public clouds or for Velero to have all uploaded backups encrypted.
The file listing with the unecrypted file size was also very crucial in my tests with different applications and this is why the only way I found was to have a suffix to decide if a file is encrypted or not without pulling the metadata.

FlorinPeter:
@FlorinPeter I was looking at the crypto/Encryption implementation and don't understand why we need to use a thread there. I expected it to wrap the supplied input stream and implements the read() method to produce the encrypted blocks (like the CipherInputStream) and padding. Can you expand on why we need the thread?

I was also not happy with that and changed it now.

Also, have you considered using the object metadata vs object suffix for the encrypted blobs? I'm curious why you settled on the suffix approach -- the object metadata seems more elegant (no changes in blob names). Do you see difficulty from switching to using the metadata?

Yes I considered a lot of possibilities and this is now my 3rd or 4th iteration.
The problem with the metadata is that on multipart uploads you need to know the metadata like size of all parts before initializing the upload this fact makes impossible to change the metadata of the concated blob with the final size and parts. I also played with content type and other things.
Right now I use the s3proxy for the docker registry to make sure that all layers are encrypted on public clouds or for Velero to have all uploaded backups encrypted.
The file listing with the unecrypted file size was also very crucial in my tests with different applications and this is why the only way I found was to have a suffix to decide if a file is encrypted or not without pulling the metadata.

docs/Encryption.md
```diff
@@ -0,0 +1,76 @@
...
+| Delimiter | 8 byte    | The delimiter is used to detect if the ```Blob``` is encrypted |
+| IV        | 16 byte   | AES initialization vector                                      |
+| Part      | 4 byte    | The part number                                                |
+| Size      | 8 byte    | The unencrypted size of the ```Blob```                         |
+| Version   | 2 byte    | Version can be used in the future if changes are necessary     |
+| Reserved  | 26 byte   | Reserved for future use                                        |
+
+### Multipart handling 
+A single ```Blob``` can be uploaded by the client into multiple parts. After the completion all parts are concatenated into a single ```Blob```.
+This procedure will result in multiple parts and paddings being held by a single ```Blob```.
```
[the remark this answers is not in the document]
FlorinPeter: This was one of the most challenging part. Because it gets really messy when you want to read from a specific offset and your blob was uploaded in multiple parts.

docs/Encryption.md
FlorinPeter: I looked in various implementations and none of them satisfice me as it comes always with some limitations that needs to taken in consideration like on AWS SDK you will need to upload the client side encrypted parts in the right order etc.

docs/Encryption.md
FlorinPeter: There is a version in the padding that can be used if it is needed in the future.

docs/Encryption.md
FlorinPeter: The cipher is ```AES/CFB/NoPadding``` and a custom padding is always being added to a blob.

src/test/java/org/gaul/s3proxy/EncryptedBlobStoreLiveTest.java
FlorinPeter: reverted my changes

docs/Encryption.md
```diff
@@ -0,0 +1,76 @@
...
+Every uploaded part get a padding of 64 bytes that includes the necessary information for decryption. The input stream from a s3 client is passed through ```CipherInputStream``` and piped to append the 64 byte part padding at the end the encrypted stream. The encrypted input stream is then processed by the ```BlobStore``` to save the ```Blob```.
+
+| Name      | Byte size | Description                                                    |
+|-----------|-----------|----------------------------------------------------------------|
+| Delimiter | 8 byte    | The delimiter is used to detect if the ```Blob``` is encrypted |
+| IV        | 16 byte   | AES initialization vector                                      |
+| Part      | 4 byte    | The part number                                                |
+| Size      | 8 byte    | The unencrypted size of the ```Blob```                         |
+| Version   | 2 byte    | Version can be used in the future if changes are necessary     |
+| Reserved  | 26 byte   | Reserved for future use                                        |
```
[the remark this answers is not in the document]
FlorinPeter: This will not work if you have for example 1000 parts or more as it will occupy the metadata space which is 2kb aur 4kb if I am not wrong. This is why each part has its own padding.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
FlorinPeter: yes, understand that but I don't see an alternative yet

timuralp:
> > Also, have you considered using the object metadata vs object suffix for the encrypted blobs? I'm curious why you settled on the suffix approach -- the object metadata seems more elegant (no changes in blob names). Do you see difficulty from switching to using the metadata?
> 
> Yes I considered a lot of possibilities and this is now my 3rd or 4th iteration. The problem with the metadata is that on multipart uploads you need to know the metadata like size of all parts before initializing the upload this fact makes impossible to change the metadata of the concated blob with the final size and parts. I also played with content type and other things. Right now I use the s3proxy for the docker registry to make sure that all layers are encrypted on public clouds or for Velero to have all uploaded backups encrypted. The file listing with the unecrypted file size was also very crucial in my tests with different applications and this is why the only way I found was to have a suffix to decide if a file is encrypted or not without pulling the metadata.

@FlorinPeter  you're right about the multipart uploads issue -- I did not consider the requirements for uploads with up to 10,000 parts. What are your thoughts on using metadata for simple objects and your current scheme for the multipart objects?

Sorry again for the delay -- I'll look through the other large object support parts this weekend.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,766 @@
...
+        LoggerFactory.getLogger(EncryptedBlobStore.class);
+    private SecretKeySpec secretKey;
+
+    private EncryptedBlobStore(BlobStore blobStore, Properties properties)
+        throws IOException {
+        super(blobStore);
+
+        String password = properties.getProperty(
+            S3ProxyConstants.PROPERTY_ENCRYPTED_BLOBSTORE_PASSWORD);
+        if (password == null || password.isEmpty()) {
```
timuralp, line 75: You can use `checkArgument(Strings.isNullOrEmpty(), "message")` like here: https://github.com/gaul/s3proxy/blob/master/src/main/java/org/gaul/s3proxy/S3Proxy.java#L74

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,766 @@
...
+        if (mbm.getUserMetadata()
+            .containsKey(Constants.METADATA_ENCRYPTION_PARTS)) {
+            int parts = Integer.parseInt(
+                mbm.getUserMetadata().get(Constants.METADATA_ENCRYPTION_PARTS));
+            int partPaddingSizes = Constants.PADDING_BLOCK_SIZE * parts;
+            long size = blobMeta.getSize() - partPaddingSizes;
+            mbm.setSize(size);
+            mbm.getContentMetadata().setContentLength(size);
+        } else {
+            // on s3 backends like aws or minio we rely on the eTag suffix
```
timuralp, line 303: I think this assumption is problematic when a server side copy is performed for objects < 5GB that were uploaded via MPU, because the new blob will not have the MPU ETag. This shows up when doing a server side copy (i.e. PutBlob with CopySource being the encrypted blob). The way around this is to convert this copy request into a new multipart upload, server-side copy the parts, then complete the upload.

I suggest failing the server side request in this case, at least to start with. That's how multipart upload server side copy has to be performed anyway.

I tested the behavior with a ~6MB object and ended up reading 2560 1024 byte chunks back after the server side copy was performed.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,766 @@
...
+
+    @Override
+    public List<MultipartUpload> listMultipartUploads(String container) {
+        List<MultipartUpload> mpus = new ArrayList<>();
+
+        // emulate list of multipart uploads on gcp
+        if (getBlobStoreType().equals("google-cloud-storage")) {
+            ListContainerOptions options = new ListContainerOptions();
+            PageSet<? extends StorageMetadata> mpuList =
+                delegate().list(container, options.prefix(".mpu/"));
```
timuralp, line 577: Use a constant for `.mpu/`

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,766 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    // filter the list by showing the unencrypted blob size
+    private PageSet<? extends StorageMetadata> filteredList(
+        PageSet<? extends StorageMetadata> pageSet) {
+        ImmutableSet.Builder<StorageMetadata> builder = ImmutableSet.builder();
+        for (StorageMetadata sm : pageSet) {
+            if (sm instanceof BlobMetadata) {
```
timuralp, line 232: Do we need to also filter out the `.mpu/` blobs in GCS? I haven't gotten around to testing this against Azure/GCS, but seemed like they would show up?

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,766 @@
...
+                    .put(Constants.METADATA_MULTIPART_KEY, mbm.getName());
+
+                // since gcp does not have a uploadId
+                // we use the sha256 of the path
+                String uploadId = generateUploadId(container, mbm.getName());
+
+                // to emulate later the list of multipart uploads
+                // we create a placeholer
+                BlobBuilder builder =
+                    blobBuilder(".mpu/" + uploadId)
```
timuralp, line 555: Use a constant for `.mpu/`

timuralp (changes requested):
I have one more high level question: can you expand on why MPU uploads on GCS an Azure Blob require special treatment (vs what S3Proxy does normally)?

Another possible issue I noticed is boundary checks for 5GB parts. However, I found a bug in S3Proxy where I can't currently upload 5GB parts, so that's moot until we fix that, but we should impose a `5GB - <number of bytes required for encryption>` as the part size limit.

FlorinPeter:
> I have one more high level question: can you expand on why MPU uploads on GCS an Azure Blob require special treatment (vs what S3Proxy does normally)?

@timuralp sorry for the delay the different treatment comes from the fact that each time a ```initiateMultipartUpload``` on GCP or Azure is called a random uuid is generated. The docker registry for example checks by listing the current uploads and try find this particular uuid but since it is random and not persisted it results in issues for the client.

> Another possible issue I noticed is boundary checks for 5GB parts. However, I found a bug in S3Proxy where I can't currently upload 5GB parts, so that's moot until we fix that, but we should impose a 5GB - <number of bytes required for encryption> as the part size limit.

Can you point me to the location where this should  be implemented or is there something like ```getMaximumMultipartPartSize``` that can be adjusted?

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
FlorinPeter: I will try to look this week into it

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,766 @@
...
+        } else {
+            // remove the -s3enc suffix while decrypting
+            // but not if it contains a multipart meta
+            if (!blobMeta.getUserMetadata()
+                .containsKey(Constants.METADATA_IS_ENCRYPTED_MULTIPART)) {
+                blobMeta = removeEncryptedSuffix(blobMeta);
+            }
+        }
+
+        // we do not set contentMD5 as it will not match due to the encryption
```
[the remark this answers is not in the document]
FlorinPeter: I would like to document this shortcoming for now if it is ok for you?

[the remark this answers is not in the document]
gaul: Honestly the Content-MD5 header is less valuable than you imagine -- I believe that the AWS SDK does not bother to calculate MD5 if either HTTPS or v4 X-Amz-Content-SHA256 are used since these provide sufficient transport-level security.  I don't think anything needs to be done here.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -1203,11 +1205,23 @@ private void handleListMultipartUploads(HttpServletRequest request,
             xml.writeEmptyElement("NextKeyMarker");
             xml.writeEmptyElement("NextUploadIdMarker");
             xml.writeEmptyElement("Delimiter");
-            xml.writeEmptyElement("Prefix");
+
+            if (prefix == null || prefix.isEmpty()) {
```
gaul, line 1209: Can call `Strings.isNullOrEmpty`.

src/main/java/org/gaul/s3proxy/crypto/DecryptionInputStream.java
```diff
@@ -0,0 +1,385 @@
...
+    private final Cipher cipher;
+
+    // the secret key
+    private final SecretKey key;
+
+    // the list of parts we expect in the stream
+    private final TreeMap<Integer, PartPadding> parts;
+
+    // the underlying input stream
+    private final InputStream input;
```
gaul, line 44: Do you need to store this?  Can you just use the `FilterInputStream` superclass `in` field?

src/test/java/org/gaul/s3proxy/EncryptedBlobStoreLiveTest.java
```diff
@@ -0,0 +1,282 @@
...
+import org.jclouds.s3.domain.ObjectMetadataBuilder;
+import org.jclouds.s3.domain.S3Object;
+import org.jclouds.s3.reference.S3Constants;
+import org.testng.SkipException;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.Test;
+
+@SuppressWarnings("UnstableApiUsage")
+@Test(testName = "EncryptedBlobStoreLiveTest")
+public final class EncryptedBlobStoreLiveTest extends S3ClientLiveTest {
```
gaul, line 56: I don't think that subclassing `S3ClientLiveTest` makes sense here -- none of the other middlewares do this.  Can you simplify this by moving the unique test coverage into `EncryptedBlobStoreTest`?

timuralp:
> > I have one more high level question: can you expand on why MPU uploads on GCS an Azure Blob require special treatment (vs what S3Proxy does normally)?
> 
> @timuralp sorry for the delay the different treatment comes from the fact that each time a `initiateMultipartUpload` on GCP or Azure is called a random uuid is generated. The docker registry for example checks by listing the current uploads and try find this particular uuid but since it is random and not persisted it results in issues for the client.

@gaul this seems like an issue in how MPUs work in AzureBlob and GCS in jclouds. I thought the provider should make a 0-sized stub for the uploads but didn't see that? Did I miss something in how jclouds works here? I'd rather fix it in jclouds than workaround it in s3proxy.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,774 @@
...
+            SecretKeyFactory factory =
+                SecretKeyFactory.getInstance("PBKDF2WithHmacSHA256");
+            KeySpec spec =
+                new PBEKeySpec(password.toCharArray(), salt.getBytes(), 65536,
+                    128);
+            SecretKey tmp = factory.generateSecret(spec);
+            secretKey = new SecretKeySpec(tmp.getEncoded(), "AES");
+        } catch (Exception e) {
+            throw new IllegalArgumentException(e);
+        }
```
fungiboletus, line 107: Why hardcoding a key derivation from a password instead of simply asking for a key ?

I would expect people able to setup s3proxy also able to generate a good key.

docs/Encryption.md
```diff
@@ -0,0 +1,76 @@
+S3Proxy
+
+# Encryption 
+
+## Motivation 
+The motivation behind this implementation is to provide a fully transparent and secure encryption to the s3 client while having the ability to write into different clouds.
+
+## Cipher mode
+The chosen cipher is ```AES/CFB/NoPadding``` because it provides the ability to read from an offset like in the middle of a ```Blob```.
```
fungiboletus, line 9: Why AES/CFB on 128 bits instead of the nowadays very common AES 256 GCM ?

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
FlorinPeter: @fungiboletus well I tried to keep it simple but I am happy if you could open a PR after it is merged that allows "expert" people to set their holy key while using a password should still be possible.

docs/Encryption.md
FlorinPeter: @fungiboletus as I wrote AES CFB allows to us to read from an offset without having to get in a worst case multiple gigabytes to only read some needed bytes.
As of now I don't saw an implementation how to read from a specific offset of a file with AES GCM but feel free to fork this PR and propose the change. Also we can handle multiple versions of the encryption in the block header so we can also change it somewhere in the future

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
```diff
@@ -0,0 +1,774 @@
...
+            .containsKey(Constants.METADATA_ENCRYPTION_PARTS)) {
+            int parts = Integer.parseInt(
+                mbm.getUserMetadata().get(Constants.METADATA_ENCRYPTION_PARTS));
+            int partPaddingSizes = Constants.PADDING_BLOCK_SIZE * parts;
+            long size = blobMeta.getSize() - partPaddingSizes;
+            mbm.setSize(size);
+            mbm.getContentMetadata().setContentLength(size);
+        } else {
+            // on s3 backends like aws or minio we rely on the eTag suffix
+            String regex = ".*-([0-9]+)";
```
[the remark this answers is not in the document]
FlorinPeter: it is not work with the trailing ``$`` because it has quotes like "db4fc7d0d5631dfb1ddfc48481243f44-3" but I will move it to ```static Pattern```

src/main/java/org/gaul/s3proxy/crypto/DecryptionInputStream.java
FlorinPeter: you are totally right

src/test/java/org/gaul/s3proxy/EncryptedBlobStoreLiveTest.java
FlorinPeter: Well, to bo honest, I know that the other middleware are not doing such a test but the idea is to see if the ```S3ProxyHandler``` is handling something different that can cause issues with the encryption.
From my point of view it does not hurt to use the standard live approach enriched with some additional tests here or?

src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java
```diff
@@ -0,0 +1,835 @@
...
+    @Test
+    public void testBlobNotEncrypted() throws Exception {
+
+        String[] tests = new String[] {
+            "1", // only 1 char
+            "123456789A12345", // lower then the AES block
+            "123456789A1234567", // one byte bigger then the AES block
+            "123456789A123456123456789B123456123456789C" +
+                "1234123456789A123456123456789B123456123456789C1234"
+        };
```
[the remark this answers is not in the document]
FlorinPeter: From my point of view it makes harder to read the code and the parameters are not everywhere the same so not much benefit as far I can see.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
FlorinPeter: this part of the code only affects the file listing functions. If you upload two parts and complete the upload then server side copy the file, it will loose the part count on the eTag but the decryption is still working as it detects the part count from the padding. But maybe it don't really get the issue, if so please help me with some s3cmd or curl commands to reproduce this behavior.

src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
fungiboletus: I understand the need to keep it simple. You do allow people to use shitty passwords though.

docs/Encryption.md
fungiboletus: GCM uses a counter of so it's relatively easy to start from an offset, you lose the authentication though. For that you need to process the whole file. For example ZFS uses AES-256-GCM by default for encryption, so it is working well in practice.

I was just curious about your choices, I'm unfortunately not very confident in crypto in Java and I see in your code that this is quite low level. It's not about changing a cipher mode in a constant somewhere 😅

Anyway your work is very appreciated 🙂

gaul:
Thank you for your contribution @FlorinPeter!  I will run a new release of S3Proxy soon.

gaul:
@FlorinPeter Could you add a wiki page and section to https://github.com/gaul/s3proxy#middlewares?  Currently this feature is not discoverable.

PR #406, opened 2022-02-10
Support configurable base directory for Docker container

liamwhite:
Allowing this to be configurable makes it much easier to deal with when symlinking in an existing filesystem-based project.


gaul:
Thank you for your contribution @liamwhite!

PR #407, opened 2022-02-10
ExAws compatibility

liamwhite:
Makes S3Proxy compatible with ExAws by testing for query string parameter _presence_ on endpoints, instead of equality to empty string.


gaul:
Thank you for your contribution @liamwhite!  It would be nice to make the ExAws library act more similarly to the REST API examples as well, even if it is not required.

PR #416, opened 2022-04-18
add Access-Control-Allow-Credentials header

### LINKED ISSUE #415: [cors] missing Access-Control-Allow-Credentials header
When trying to access a file using a browser and a cors request that includes cookies, the request fails due to:

```Credentials flag is true, but Access-Control-Allow-Credentials is not "true".```

This is a browser limitation. Example of a fetch request:

```
fetch(url, {
    credentials: 'include',
    mode: 'cors'
})
```

Pull request: https://github.com/gaul/s3proxy/pull/416

st-h:
fixes #415


gaul:
@reimannf could you review this?

reimannf:
I think CORS does not include cookies/credentials on cross-origin requests by default. Both, client and server has to opt-in if they want to do this. So instead of always sending that header, which might open some attacking space, I suggest to introduce a config property `PROPERTY_CORS_ALLOW_CREDENTIAL` here: https://github.com/gaul/s3proxy/blob/217308abd735c5f6e31ef4cf8351d5b287a5dc03/src/main/java/org/gaul/s3proxy/S3ProxyConstants.java#L41-L42

You might want to send the header always with the value configured in `PROPERTY_CORS_ALLOW_CREDENTIAL` unless you default to something different than `true` and make this an active decision.

reimannf:
I think CORS does not include cookies/credentials on cross-origin requests by default. Both, client and server has to opt-in if they want to do this. So instead of always sending that header, which might open some attacking space, I suggest to introduce a config property `PROPERTY_CORS_ALLOW_CREDENTIAL` here: https://github.com/gaul/s3proxy/blob/217308abd735c5f6e31ef4cf8351d5b287a5dc03/src/main/java/org/gaul/s3proxy/S3ProxyConstants.java#L41-L42

You might want to send the header always with the value configured in `PROPERTY_CORS_ALLOW_CREDENTIAL` unless you default to something different than `true` and make this an active decision.

gaul:
@st-h Could you address the review comment so I can merge this?

st-h:
@gaul thanks for the reminder. I totally forgot this PR was still open and I was chugging along with my local docker image. Would have been up for a surprise when switching to a different computer. I just added the request property.

gaul:
Thank you for your contribution and patience @st-h!

Issue #404, opened 2022-02-07
How to run docker against Azure Storage?


AArnott:
I'm trying to piece together from the wiki and the dockerfile how to run the s3proxy against an Azure Blob Storage account. This is what I have:

```
docker run \
    --publish 80:80 \
    --env S3PROXY_AUTHORIZATION=none \
    --env JCLOUDS_PROVIDER=azureblob \
    --env JCLOUDS_ENDPOINT=arnottmedia.core.windows.net \
    --env JCLOUDS_IDENTITY=arnottmedia \
    --env JCLOUDS_CREDENTIAL=9qaV... \
    andrewgaul/s3proxy
```

But this fails when accessing http://localhost:80 with this error:

> HTTP ERROR 400 endpoint.getHost() is null for /arnottmedia.core.windows.net/?comp=list&amp;include=metadata

I tried with the endpoint set to just `arnottmedia` as well, to no avail. I don't know what to put for identity because Azure only offers 'keys' (no IDs). 

Any suggestions?

timuralp:
There are two ways to configure Azure Blob access. One possibility is using the Azure tenant ID, client ID, and password. The settings would look like:
```
jclouds.provider=azureblob
jclouds.identity=<client ID>
jclouds.credential=<password>
jclouds.azureblob.tenantId=<tenand ID>
jclouds.azureblob.account=<storage account>
jclouds.azureblob.auth=azureAd
```

The other method is via Azure storage Keys, which uses the key and storage account name and looks like:
```
jclouds.provider=azureblob
jclouds.credential=<key>
jclouds.identity=<storage account>
```

I don't think you need to specify the endpoint -- S3Proxy should infer it from the storage account name that was passed in.

AArnott:
Thanks. I tried the latter, and it still gives me that same getHost error message

```
docker run \
    --publish 80:80 \
    --env S3PROXY_AUTHORIZATION=none \
    --env JCLOUDS_PROVIDER=azureblob \
    --env JCLOUDS_IDENTITY=arnottmedia \
    --env JCLOUDS_CREDENTIAL=9qaV... \
    andrewgaul/s3proxy
```

timuralp:
@AArnott Thanks for testing it out. Looks like the `-Djclouds.endpoint=""` is not handled properly by the underlying jclouds library, resulting in an empty string for the endpoint and hence the above error. Two workarounds until the jclouds issue is fixed:
[...]

PR #420, opened 2022-06-20
Docker entrypoint: authentication for Azure Blob

christophkreutzer:
When using s3proxy as a proxy for S3 -> Azure Blob Storage, one needs to set the authentication parameters. When running in a container, these can be exposed via environment variables, but are currently not passed through to the jclouds config.

This is how we got it running for us in a custom build of the container.
Should I also add the options for other backends? Or do you see options for a generic solution?

Maybe also corresponds to #404


gaul:
Thank you for your contribution @christophkreutzer!  I am happy to add configs as one-offs but if there is some better mechanism please suggest it.

christophkreutzer:
Hello @gaul, thanks for merging. I have no clear idea about a generic implementation, at least with env variables.

Another way would be allow passing a properties file (which could be Docker file mount or a configmap mount on Kubernetes), but that's a bit less elegant to use in my opinion.

PR #426, opened 2022-06-30
feat!: Increase minimum Java version

ryanfaircloth:
Increase the minimum Java version to allow for update of Jetty


gaul:
Not sure what this accomplishes but #457 supersedes it.

PR #441, opened 2022-07-19
Allow setting of v4-max-non-chunked-request-size in Docker container

thiagodasilva:


gaul:
Thank you for your contribution @thiagodasilva!  Could you share why you want to tune this?  I worry that the default value is not correct.  I picked the current one when testing against AWS several years ago.

thiagodasilva:
@gaul thanks for merging so quickly. When I was testing uploading MPUs with different part sizes, I wasn't able to upload anything larger than 32MB. IIUC, MPU parts are allowed to be as large as 5 GBs?

PR #505, opened 2023-03-17
Add S3PROXY_JAVA_OPTS environment variable

srujandeshpande:
Add a `S3PROXY_JAVA_OPTS` environment variable into the run-docker-container.sh file. This will allow users to easily pass in custom options to the JVM. 

For example, my use case is to pass `-Xmx` and `-Xms` which can be passed using this variable

### DIFF
--- a/src/main/resources/run-docker-container.sh
+++ b/src/main/resources/run-docker-container.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 exec java \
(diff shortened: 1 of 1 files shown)


srujandeshpande:
@gaul, could you please review this?

gaul:
Thank you for your contribution @srujandeshpande!

PR #512, opened 2023-04-18
Configure Keystore through environment variables

kinoute:
The documentation about SSL support explains that we need to change the endpoint, the keystore path and the keystore password. But only the endpoint is actually configurable via environment variables when we use Docker.

This PR allows configuration of the keystore path and keystore password through environment variables.


gaul:
Thank you for your contribution @kinoute!

Issue #479, opened 2022-12-26
CI failing


gaul:
Looks like some rot within s3-tests:

> AttributeError: module 'collections' has no attribute 'Callable'

Seen in #464.

gaul:
@timuralp the Containerize steps fails with:

```
 Failed to get manifest docker.io/library/master:latest: Unauthorized [http 401]: {"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":[{"Type":"repository","Class":"","Name":"library/master","Action":"pull"}]}]}
```

https://github.com/gaul/s3proxy/actions/runs/5020361449/jobs/9001824708#step:10:29

Do you think this is an easy fix or should we just disable the Docker push?

timuralp:
I need to investigate this a bit. Will try to look into this tonight/tomorrow. I don't understand why the login doesn't work.

PR #514, opened 2023-04-22
Update s3-tests submodule

gaul:
References #479.


gaul:
This includes gaul/s3-tests#1 from @kinoute that may address the recent CI failures.

gaul:
This does address the original symptom although it looks like a few tests regressed during the CI breakage.

PR #540, opened 2023-08-16
explicitly depend on plexus-utils in modernizer plugin

larshagencognite:
The implicit dependency on plexus-utils was dropped in maven 3.9.0, so this fix is needed to build on newer versions of maven.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -343,6 +343,13 @@
         <configuration>
           <javaVersion>${java.version}</javaVersion>
         </configuration>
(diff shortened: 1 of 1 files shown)


larshagencognite:
In that case it should be fixed by https://github.com/gaul/modernizer-maven-plugin/commit/35596dc589b0bccf20573ca0c78fe5e23a09052b

gaul:
Sure please open a PR for this.

PR #542, opened 2023-08-16
update modernizer plugin to 2.6.0

larshagencognite:
This allows the project to work with maven 3.9+, as the modernizer no longer implicitly depends on plexus-utils.


gaul:
Why not upgrade to 2.6.0?

larshagencognite:
That was just being conservative, happy to change it to 2.6.0

gaul:

PR #543, opened 2023-08-17
make sure all errors are logged at debug level

larshagencognite:
Printing a stack trace for failing requests is useful for debugging, and for HEAD requests it is even more important, as there is no response body with an error message.


gaul:
This makes sense and matches other unknown cases.  Thank you for your contribution @larshagencognite!

PR #557, opened 2023-10-01
Fix port out of range in container

### LINKED ISSUE #556: Version 2.1 in Docker container fails with port out of range
When properties are passed to java via -D they are empty strings rather than null this is causing the container to fail to start

ryanfaircloth:
fixes #556, check that endpoint strings are not null and not empty

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3Proxy.java
+++ b/src/main/java/org/gaul/s3proxy/S3Proxy.java
@@ -31,6 +31,9 @@ import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import org.eclipse.jetty.http.HttpCompliance;
 import org.eclipse.jetty.server.HttpConfiguration;
 import org.eclipse.jetty.server.HttpConnectionFactory;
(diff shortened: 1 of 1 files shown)


src/main/java/org/gaul/s3proxy/S3Proxy.java
```diff
@@ -167,10 +175,12 @@ public static Builder fromProperties(Properties properties)
                         S3ProxyConstants.PROPERTY_ENDPOINT + " or " +
                         S3ProxyConstants.PROPERTY_SECURE_ENDPOINT);
             }
-            if (endpoint != null) {
+            if (endpoint != null && !endpoint.isEmpty()) {
```
gaul, line 178: Can you call `!Strings.isNullOrEmpty(endpoint)` instead?

src/main/java/org/gaul/s3proxy/S3Proxy.java
```diff
@@ -145,6 +151,8 @@ public static final class Builder {
         private CrossOriginResourceSharing corsRules;
         private int jettyMaxThreads = 200;  // sourced from QueuedThreadPool()
         private int maximumTimeSkew = 15 * 60;
+        private static final Logger logger = LoggerFactory.getLogger(
```
gaul, line 154: I think it should be OK to use the `S3Proxy` logger you added above.  But you might need to remove the `private` access modifier so it is package-protected.

gaul (changes requested):
Please fix small nits.

gaul:
Superseded by #570.

PR #561, opened 2023-10-11
check NoSuchFileException for handler

jixinchi:
When using filesystem, I found that OPTIONS a non-exist bucket will cause exceptions and returns 500. I found the reason is that jcloud.blobstore does not check file exists in getContainerAccess, and not throws ContainerNotFoundException.


gaul:
Should this be fixed in jclouds instead?

jixinchi:
> Should this be fixed in jclouds instead?

Good idea! I'll see how to contribute to jclouds

PR #563, opened 2023-10-12
upgrade OS packages in Dockerfile

larshagencognite:
This reduces the number of vulnerabilities in the final image

Before (reported by trivy)
Total: 141 (UNKNOWN: 0, LOW: 68, MEDIUM: 32, HIGH: 36, CRITICAL: 5)
After (reported by trivy)
Total: 89 (UNKNOWN: 0, LOW: 68, MEDIUM: 10, HIGH: 10, CRITICAL: 1)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,12 @@
 FROM openjdk:11-jre-slim
 LABEL maintainer="Andrew Gaul <andrew@gaul.org>"
 
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
 WORKDIR /opt/s3proxy
 
 COPY \


gaul:
Is this the correct approach?  Shouldn't this image be periodically updated?  Instead should we update to [openjdk:17.0.1-jdk-slim](https://hub.docker.com/layers/library/openjdk/17.0.1-jdk-slim/images/sha256-565d3643a78a657ca03e85c110af9579a07e833d6bcc14f475249c521b5c5d74) or preferably [openjdk:21-jdk-slim](https://hub.docker.com/layers/library/openjdk/21-jdk-slim/images/sha256-58b6a71935e33dc39785423027a6f957c749f74eeb90b7937587c27fa4f5cab2)?

larshagencognite:
We should not continue to use openjdk, as it is deprecated, I have another PR to start moving to eclipse-temurin https://github.com/gaul/s3proxy/pull/564

gaul:
I'm happy to run `apt-get upgrade` as well but I'd like to understand why stale images are being published.  I would think that the maintainers of these should regularly update them to avoid security issues?

larshagencognite:
I did a spot check just now, and the latest version of `eclipse-temurin:21-jre` did not contain any vulnerabilities that could be fixed by upgrades. It did not contain any HIGH or CRITICAL severity vulnerabilities (in trivy terms).

So I agree that we should be able to trust the maintainers to keep the image up to date, and rather revisit this later if there is a need.

gaul:
Thanks for investigating this!  Can you submit a PR to upgrade to `eclipse-temurin:21-jre`?

PR #565, opened 2023-10-13
Add DELETE and HEAD methods for CORS and CORS headers for all responses

jixinchi:
1. I found that DELETE method is not supported for CORS, which should be included. Refer to: [aws CORS config](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ManageCorsUsing.html#cors-allowed-methods)

2. Currently, not all handle methods include "addCorsResponseHeader". I add it in all handle methods, because in most cases, the browser needs it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ s3proxy.cors-allow-credential=true
 ```
 
 CORS cannot be configured per bucket. `s3proxy.cors-allow-all=true` will accept any origin and header.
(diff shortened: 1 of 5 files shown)


gaul:
@reimannf could you look at this?

reimannf:
LGTM. Only one nitpick to update the README here: https://github.com/gaul/s3proxy/blob/master/README.md?plain=1#L148

gaul:
Thank you for your contribution @jixinchi!

PR #569, opened 2023-10-24
Restore OPAQUE_MARKERS handling for list objects V1

larshagencognite:
We need this to support V1 clients that use the last object name in the response as a marker for the next page (which is allowed according to the S3 docs).

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -1490,8 +1490,12 @@ public class S3ProxyHandler {
                 if (Quirks.OPAQUE_MARKERS.contains(blobStoreType)) {
                     StorageMetadata sm = Streams.findLast(set.stream()).orElse(null);
                     if (sm != null) {
-                        lastKeyToMarker.put(Maps.immutableEntry(containerName,
-                                encodeBlob(encodingType, nextMarker)), nextMarker);
+                        // TODO: verify if we need this handling at all for V2
+                        String lastKey = isListV2 ? encodeBlob(encodingType, nextMarker) : sm.getName();
+                        lastKeyToMarker.put(
+                                Maps.immutableEntry(containerName, lastKey),
+                                nextMarker
+                        );
                     }
                 }
             } else {


larshagencognite:
@Decard6 This partially reverts changes in https://github.com/gaul/s3proxy/pull/531

gaul:
I'm concerned that s3-tests isn't exercising this behavior.  Can you look at these and see what's missing?

larshagencognite:
Do we have automated testing against a real azure blob store?

larshagencognite:
@gaul Could you give me some pointers to where this should be tested?

gaul:
The ideal way to test this is https://github.com/ceph/s3-tests which I periodically rebase onto.  But I don't have any automated integration tests against Azure presently.

Alternatively maybe we could parameterize some S3Proxy listObjects tests to toggle `OPAQUE_MARKERS`?  I'm happy to merge this PR but just want to make sure that we have some way to avoid future regressions.

gaul:
@larshagencognite could you rebase this change?

gaul:
@Decard6 any thoughts here?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -1490,8 +1490,12 @@ private void handleBlobList(HttpServletRequest request,
                 if (Quirks.OPAQUE_MARKERS.contains(blobStoreType)) {
                     StorageMetadata sm = Streams.findLast(set.stream()).orElse(null);
                     if (sm != null) {
-                        lastKeyToMarker.put(Maps.immutableEntry(containerName,
-                                encodeBlob(encodingType, nextMarker)), nextMarker);
+                        // TODO: verify if we need this handling at all for V2
+                        String lastKey = isListV2 ? encodeBlob(encodingType, nextMarker) : sm.getName();
```
Decard6, line 1494: @gaul 
As long as `isListV2` works as intended it should be good. Regardless I'd want to test the changes.

Decard6: All good. The change doesn't break pagination.

jdunham-openai:
What do you need to test (I can) to merge this?  I confirmed that v1 list works after this change as I was having issues.

gaul:
It would be great if there were some kind of test that exercises this condition so I can confidently merge this PR.  Could you submit a test that fails before this PR and succeeds with it?  Note that CI includes the Azurite storage emulator so this should be easy.

gaul:
I'm preparing a new release so please try to address my comments before the weekend.

gaul:
Going with a different approach please test the latest master.

gaul:
#938 will complete this work.

PR #570, opened 2023-10-24
treat null and empty string equally for endpoint properties

larshagencognite:
This improves the usability of the docker image, as it defaults to empty strings for both properties. With this fix, the docker image will work when only one of the endpoints are set.


larshagencognite:
Fixes https://github.com/gaul/s3proxy/issues/523, https://github.com/gaul/s3proxy/issues/528 and https://github.com/gaul/s3proxy/issues/556

gaul:

PR #574, opened 2023-11-13
commonPrefixes is only for delimiter request; check size and tier null

jixinchi:


gaul:
These seem to make sense but could you explain the user symptoms more clearly, e.g., which provider does this currently fail on?

jixinchi:
> These seem to make sense but could you explain the user symptoms more clearly, e.g., which provider does this currently fail on?

I'm using `filesystem` as provider, and I have a problem of deleting a "directory" recursively by aws cli. The cli list files with `Prefix` and without `Delimeter`, and then delete each items in the `Contents` field of response. If there are empty directories below the directory to be deleted, these empty directories can not be deleted because they are not in the `Contents` filed.

jixinchi:
Sorry, there are something wrong in above expaination. If the request does not contains delimeter, it should be added to `Contents`. So I `break` the switch-case, not `continue` to next item

gaul:
Thank you for your contribution and explanation @jixinchi!

Issue #591, opened 2024-01-09
support 32-bit arm on dockerhub?


snpz:
Hi.

I'm using a rather old [Wandboard Quad](https://eu.mouser.com/ProductDetail/Wandboard/WB-IMX6QP-BW?qs=Mv7BduZupUiCXDZfcox9kA%3D%3D) which only has a 32-bit processor.
I would like to run s3proxy via docker, but currently there is only arm64 available, no 32-bit version.

Without docker it works without issues with openjdk 11.
```
# java -version
# openjdk version "11.0.21" 2023-10-17
# OpenJDK Runtime Environment (build 11.0.21+9-post-Debian-1deb11u1)
# OpenJDK Server VM (build 11.0.21+9-post-Debian-1deb11u1, mixed mode)
```

I found #382 and it made its way into dockerhub too. Would it be possible to also provide this for my kind of outdated cpu-architecture?
```
# uname -a
# Linux wandboard 5.10.0-26-armmp #1 SMP Debian 5.10.197-1 (2023-09-29) armv7l GNU/Linux
```

I'd totally understand if not. :)

gaul:
Currently the automated Docker push via CI is broken #479.  Before a contributor added ARM64 support, this worked for x86-64.  Ideally I would like to repair this if someone can tell me what is broken.  Then I would be happy to accept a PR to include 32-bit ARM as part of the automated push.  I don't have much background in Docker so this will likely need to be a contribution from the community.

PR #592, opened 2024-01-10
fix pushing images to docker.io

### LINKED ISSUE #479: CI failing

snpz:
Hi @gaul.

I took me a lot of time, reading and trial and error but I've finally come up with something to fix #479 and prepare for issue #591 I created myself yesterday.

I never used Github Actions myself so I googled a bit and found out that the [official action](https://docs.docker.com/build/ci/github-actions/push-multi-registries/) that is used to push to ghcr.io can also be used to push to docker.io without relying on regctl.

To do this one needs to pass [multiple image names](https://docs.docker.com/build/ci/github-actions/manage-tags-labels/) that are the base for building the targets to push to.

According to how the tags in the meta action are currently setup these changes resulted in 3 tags getting pushed to both ghcr.io and docker.io each time I pushed to my branch. (docker.io depending on the existence of the DOCKER_PASS secret as before)

## docker.io

docker.io

## ghrc.io
ghrc.io

I also put the variable that checks for presence of the secret on the top level as it made it more readable and use some never action versions in a few places. And configured caching via [github actions](https://docs.docker.com/build/cache/backends/gha/#using-dockerbuild-push-action). This cache type is still experimental but the recommended one.

I'm not contributing a lot so if I miss something that is expected please let me know.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -6,34 +6,33 @@ on:
   pull_request:
     branches:
       - "*"
+
+env:
+  dockerhub_publish: ${{ secrets.DOCKER_PASS != '' }}
+
 jobs:
+  
   meta:
     runs-on: ubuntu-latest
     outputs:
-      dockerhub-publish: ${{ steps.dockerhub-publish.outputs.defined }}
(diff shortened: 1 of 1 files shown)


gaul:
Let's see if this improves things.  Thanks for your contribution @snpz!

Issue #637, opened 2024-05-23
building in CI with podman fails with: "short-name resolution enforced but cannot prompt without a TTY"


vilmosnagy:
Building with podman in a CI env fails with the following message: short-name resolution enforced but cannot prompt without a TTY

PR #638, opened 2024-05-23
#637 use full names for base images

vilmosnagy:


gaul:
Thank you for your contribution @vilmosnagy!

Issue #648, opened 2024-06-10
I get a RequestTimeTooSkewed error when trying to get a file using a pre-signed url


timursaikaliev:
It looks like the isTimeSkewed check is not working properly.

```
[s3proxy] D 06-10 15:51:45.496 S3Proxy-Jetty-18 o.gaul.s3proxy.S3ProxyHandler:301 |::] request: Request(GET http://....zip?X-Amz-Expires=604800&X-Amz-Date=20240610T103411Z&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=AKIA......%2F20240610%2Feu-central-1%2Fs3%2Faws4_request&X-Amz-SignedHeaders=host&X-Amz-Signature=e37cb37c47f66c84b2ae675ba6cc65b854f74a687984fcb7059a553f7fee6b08)@10244eb9
[s3proxy] D 06-10 15:51:45.497 S3Proxy-Jetty-18 o.gaul.s3proxy.S3ProxyHandler:2935 |::] time skewed 1718015651 1718034705
[s3proxy] D 06-10 15:51:45.497 S3Proxy-Jetty-18 o.gaul.s3proxy.S3ProxyHandler:2953 |::] sendSimpleErrorResponse: 403 RequestTimeTooSkewed Forbidden {}
```

Since there is X-Amz-Expires=604800, the url should work for 7 days, but url works only 15 minutes after creation.
After that I get an error: **RequestTimeTooSkewed**.
The date on the client and server are correct.
Directly from amazon s3 everything works as it should.

Maybe I have misconfigured something? 
Please help me with a solution to the problem. Thanks.

timursaikaliev:
Also could you please add the ability to configure **s3proxy.maximum-timeskew** when running the docker image.

timursaikaliev:
According to [this,](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationTimeStamp) "The validation constraint on request date applies only to authenticated requests that do not use query string authentication. "

PR #649, opened 2024-06-11
Configure maximum-timeskew with environment variable

timursaikaliev:
Adding the ability to configure s3proxy.maximum-timeskew using an environment variable.


gaul:
Ignoring s3-tests dependency installation failure.  References #648.  Thank you for your contribution @timursaikaliev!

PR #651, opened 2024-06-12
Remove timestamp validation for requests that use query string authentication

timursaikaliev:
Relates to #648

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -447,9 +447,6 @@ public class S3ProxyHandler {
                     dateSkew = parseIso8601(request.getHeader(
                             AwsHttpHeaders.DATE));
                 }
-            } else if (request.getParameter("X-Amz-Date") != null) { // v4 query
-                String dateString = request.getParameter("X-Amz-Date");
-                dateSkew = parseIso8601(dateString);
             } else if (hasDateHeader) {
                 try {
                     dateSkew = request.getDateHeader(HttpHeaders.DATE);


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -447,9 +447,6 @@ public final void doHandle(HttpServletRequest baseRequest,
                     dateSkew = parseIso8601(request.getHeader(
                             AwsHttpHeaders.DATE));
                 }
-            } else if (request.getParameter("X-Amz-Date") != null) { // v4 query
-                String dateString = request.getParameter("X-Amz-Date");
-                dateSkew = parseIso8601(dateString);
```
gaul, line 452: Could you add a unit test that shows this problem more clearly?  This will allow fixing the underlying issue more easily.

gaul (changes requested):

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
gaul: Sorry I just saw https://github.com/gaul/s3proxy/issues/648#issuecomment-2163760808.  I'm trying to understand the intention from AWS -- do the `Date` parameters/headers only apply to unauthenticated requests and the `Expires` parameters apply to authenticated requests?

timursaikaliev: - **Date Header:** Used in both unauthenticated and authenticated requests. For authenticated requests, it's part of the signing process.

- **Expires Parameter:** Used for pre-signed URLs to specify how long the URL is valid. It applies to authenticated requests where temporary access is provided.

When using pre-signed URLs in AWS, the `Date` header is generally not required as the timestamp and expiration are included in the URL parameters themselves. The crucial components in a pre-signed URL are the `Expires` parameter and the signature, which are used to validate the request.

Even if the `Date` header is not included, AWS validates the pre-signed URL based on the following:

1. **Expiration Check:**

- AWS checks the current time against the `Expires` parameter in the URL. If the current time is beyond the expiration time specified, the request is denied.

2. **Signature Verification:**

- AWS verifies that the signature is correct by recalculating the expected signature using the provided parameters (including `Expires`).

- This ensures the URL has not been altered since it was generated.

timursaikaliev: So, if we use a pre-signed URL with no Date header, but even with Date parameters in the request, we don't need to validate the timestamp,  because if we do, we will always get an error when maximumTimeSkew expires, which is 15 minutes by default.

gaul:
Thank you for your contribution @timursaikaliev!  Does anything inside of s3-test exercise this that we could enable?

PR #666, opened 2024-07-20
Add middleware to override storage class.

### LINKED ISSUE #625: Create a middleware to change the storageclass
Some tools are configured to write to s3 with no option to configure the [storageclass](https://aws.amazon.com/s3/storage-classes/). This feature request will let those tools use the proxy with default or no storageclass and the proxy will write to the proxied s3 with a target storageclass.

We need that because some tools predates the creation of the storageclass feature brought up by AWS. We also have case where these tools has this option for AWS but not for S3 compatible, however [minio](https://github.com/minio/minio/blob/master/docs/erasure/storage-class/README.md) or [ceph](https://docs.ceph.com/en/latest/radosgw/placement/) let you build s3 solution with different storage class.

The natural approach would be to change those tools to add the storageclass information in their PUT request but often this feature is not prioritised by the development team for different reasons like upgrading their library create too many frictions with their existing codebase.

Here an example of how one would use this middleware 
```
s3proxy.storageclass.default=STANDARD_IA
s3proxy.storageclass.STANDARD=STANDARD_IA
```

this configuration make sure that no information on the storageclass (default) or the information STANDARD storageclass is provided then s3proxy will write to the STANDARD_IA storageclass.

In this situation 
```
s3proxy.storageclass.GLACIER=HDD_SC
```
Any request made for putting the object on GLACIER will be translated in HDD_SC.

This request should also work the other way, any GET request that return HDD_SC should be transformed in GLACIER .

gaul:
This is best-effort and some storage classes do not map exactly, particularly for non-S3 object stores.  Fixes #625.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/Main.java
+++ b/src/main/java/org/gaul/s3proxy/Main.java
@@ -56,6 +56,7 @@ import org.jclouds.concurrent.config.ExecutorServiceModule;
 import org.jclouds.location.reference.LocationConstants;
 import org.jclouds.logging.slf4j.config.SLF4JLoggingModule;
 import org.jclouds.openstack.swift.v1.blobstore.RegionScopedBlobStoreContext;
+import org.jclouds.s3.domain.ObjectMetadata.StorageClass;
 import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
@@ -269,6 +270,20 @@ public final class Main {
             System.err.println("Using encrypted storage backend");
             blobStore = EncryptedBlobStore.newEncryptedBlobStore(blobStore,
                 properties);
+        }
+
+        var storageClass = properties.getProperty(
+                S3ProxyConstants.PROPERTY_STORAGE_CLASS_BLOBSTORE);
+        if (!Strings.isNullOrEmpty(storageClass)) {
+            System.err.println("Using storage class override backend");
+            var storageClassBlobStore =
+                    StorageClassBlobStore.newStorageClassBlobStore(
+                            blobStore, storageClass);
+            blobStore = storageClassBlobStore;
+            System.err.println("Configuration storage class: " + storageClass);
+            // TODO: This only makes sense for S3 backends.
+            System.err.println("Mapping storage storage class to: " +
+                    StorageClass.fromTier(storageClassBlobStore.getTier()));
         }
 
         return blobStore;
--- a/src/main/java/org/gaul/s3proxy/StorageClassBlobStore.java
+++ b/src/main/java/org/gaul/s3proxy/StorageClassBlobStore.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2014-2021 Andrew Gaul <andrew@gaul.org>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gaul.s3proxy;
+
+import org.jclouds.blobstore.BlobStore;
+import org.jclouds.blobstore.domain.Blob;
+import org.jclouds.blobstore.domain.BlobMetadata;
+import org.jclouds.blobstore.domain.MultipartUpload;
+import org.jclouds.blobstore.domain.Tier;
+import org.jclouds.blobstore.domain.internal.BlobMetadataImpl;
+import org.jclouds.blobstore.options.PutOptions;
+import org.jclouds.blobstore.util.ForwardingBlobStore;
+import org.jclouds.s3.domain.ObjectMetadata.StorageClass;
+
+/**
+ * This class implements a middleware to set the storage tier when creating
+ * objects.  The class is configured via:
+ *
+ *   s3proxy.storage-class-blobstore = VALUE
+ *
+ * VALUE can be anything from org.jclouds.s3.domain.StorageClass, e.g.,
+ * STANDARD, STANDARD_IA, GLACIER_IR, DEEP_ARCHIVE.  Some values do not
+ * translate exactly due to jclouds limitations, e.g., REDUCED_REDUNDANCY maps
+ * to STANDARD.  This mapping is best effort especially for non-S3 object
+ * stores.
+ */
+public final class StorageClassBlobStore extends ForwardingBlobStore {
+    private final Tier tier;
+
+    private StorageClassBlobStore(BlobStore delegate,
+            String storageClassString) {
+        super(delegate);
+        StorageClass storageClass;
+        try {
+            storageClass = StorageClass.valueOf(
+                    storageClassString.toUpperCase());
+        } catch (IllegalArgumentException iae) {
+            storageClass = StorageClass.STANDARD;
+        }
+        this.tier = storageClass.toTier();
+    }
+
+    static StorageClassBlobStore newStorageClassBlobStore(BlobStore blobStore,
+            String storageClass) {
+        return new StorageClassBlobStore(blobStore, storageClass);
+    }
+
+    public Tier getTier() {
+        return tier;
+    }
+
+    @Override
+    public String putBlob(String containerName, Blob blob) {
+        var newBlob = replaceTier(containerName, blob);
+        return delegate().putBlob(containerName, newBlob);
+    }
+
+    @Override
+    public String putBlob(String containerName, Blob blob,
+            PutOptions options) {
+        var newBlob = replaceTier(containerName, blob);
+        return delegate().putBlob(containerName, newBlob, options);
+    }
+
+    @Override
+    public MultipartUpload initiateMultipartUpload(
+            String container, BlobMetadata blobMetadata, PutOptions options) {
+        var newBlobMetadata = replaceTier(blobMetadata);
+        return delegate().initiateMultipartUpload(container, newBlobMetadata,
+                options);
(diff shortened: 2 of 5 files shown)


gaul:
@michaelcourcy could you take a look at this to see if it meets your needs?

michaelcourcy:
Hi @gaul, is there any reason that you chose the word `tier` instead of `storageClass` in the way you name your constant and class ? 
```
public static final String PROPERTY_TIER_BLOBSTORE = "s3proxy.tier-blobstore";
public final class TierBlobStore extends ForwardingBlobStore
```
Maybe tier relates to something else, something more global that make sense in the context of s3proxy ?

michaelcourcy:
Ok I understand it relates to `org.jclouds.blobstore.domain.Tier`

gaul:
Using `Tier` is probably confusing to users -- perhaps the user-facing interface should only reference S3 concepts?  I imagine most users will use S3-S3 mapping although S3-other object store will probably require understanding the jclouds concepts.  BTW the jclouds `ARCHIVE` tier actually maps to [S3 deep archive storage class](https://github.com/apache/jclouds/blob/f4e986f8fb8b48bb3b14ce2b9ed075462f2b552f/apis/s3/src/main/java/org/jclouds/s3/domain/ObjectMetadata.java#L52).  Unfortunately there will be some lossy conversions unless we add further support to jclouds.

michaelcourcy:
> BTW the jclouds ARCHIVE tier actually maps to [S3 deep archive storage class](https://github.com/apache/jclouds/blob/f4e986f8fb8b48bb3b14ce2b9ed075462f2b552f/apis/s3/src/main/java/org/jclouds/s3/domain/ObjectMetadata.java#L52). 

Ah I missed the mapping code 
```
         case STANDARD: return StorageClass.STANDARD;
         case INFREQUENT: return StorageClass.STANDARD_IA;
         case COOL: return StorageClass.STANDARD_IA;
         case COLD: return StorageClass.GLACIER_IR;
         case ARCHIVE: return StorageClass.DEEP_ARCHIVE;
```

> Unfortunately there will be some lossy conversions unless we add further support to jclouds.

It's already a lot and far enough for starting this new feature IMO. 

I will try to test it against AWS S3 and then later against CEPH. It's been a long time I haven't build a maven project.

gaul:
Thinking about this some more, anything that requires reading the source code to be a non-starter.  I think this would work better if the user specifies the desired storage class in S3 terms and S3Proxy should log when it does a lossy conversion.  If we round-trip the StorageClass -> Tier -> StorageClass we have lossless conversions for:

```
      STANDARD(Tier.STANDARD),
      STANDARD_IA(Tier.COOL),
      GLACIER_IR(Tier.COLD),
      DEEP_ARCHIVE(Tier.ARCHIVE);
```

And lossy conversions for:

```
      INTELLIGENT_TIERING(Tier.STANDARD),
      REDUCED_REDUNDANCY(Tier.STANDARD),
      ONEZONE_IA(Tier.COOL),
      GLACIER(Tier.ARCHIVE),
```

Logging the conversions makes even more sense for non-S3 object stores.  But I imagine most users will use S3 and be happy with the lossless mapping of the first four storage classes.

michaelcourcy:
> I think this would work better if the user specifies the desired storage class in S3 terms
If user proxy against a s3 for sure but if he proxies against Azure or GCP then this s3 mapping may be confusing and against the spirit of the project ? 

> anything that requires reading the source code is a non-starter. 
Sure but we'll add a wiki page

michaelcourcy:
@gaul My test fails but maybe I missed the configuration.

I checkout tier-blobstore branch and build
```
mvn package
```
s3proxy.properties I introduce `s3proxy.tier-blobstore=INFREQUENT`
```
s3proxy.tier-blobstore=INFREQUENT
s3proxy.endpoint=http://127.0.0.1:8080
s3proxy.authorization=aws-v2-or-v4
s3proxy.identity=local-identity
s3proxy.credential=local-credential 
jclouds.provider=aws-s3
jclouds.identity=<redacted>
jclouds.credential=<redacted>
```
launch s3proxy `./s3proxy --properties s3proxy.properties`

then check my configuration works 
```
export AWS_ACCESS_KEY_ID=local-identity
export AWS_SECRET_ACCESS_KEY=local-credential
aws s3 --endpoint-url http://127.0.0.1:8080  ls s3://mcourcy-testia/
                           PRE k10/
                           PRE mic/
```

And try to put a file 
```
echo "put me in test" > test.txt
aws s3 --endpoint-url http://127.0.0.1:8080 cp test.txt s3://mcourcy-testia/test.txt
```

But when I check the storageclass of test.txt it's STANDARD instead of STANDARD_IA
```
aws  s3api head-object --endpoint-url http://127.0.0.1:8080 --bucket mcourcy-testia --key test.txt
```
I get 
```
aws  s3api head-object --endpoint-url http://127.0.0.1:8080 --bucket mcourcy-testia --key test.txt
{
    "LastModified": "2024-07-21T14:52:10+00:00",
    "ContentLength": 15,
    "ETag": "\"a99a4b4b9e8204c1489f9f37dd21ec1f\"",
    "ContentEncoding": "",
    "ContentType": "text/plain",
    "Metadata": {},
    "StorageClass": "STANDARD"
}
```

I confirm by checking in AWS console 
image

gaul:
Sorry please try the latest proposed PR -- previously it lacked the changes to Main.java to actually accept the configuration.

michaelcourcy:
It works now ! 
Here is the configuration I used 
```
s3proxy.endpoint=http://127.0.0.1:8080
s3proxy.authorization=aws-v2-or-v4
s3proxy.identity=local-identity
s3proxy.credential=local-credential
s3proxy.tier-blobstore=INFREQUENT
jclouds.provider=aws-s3
jclouds.identity=<redacted>
jclouds.credential=<redacted>
```

michaelcourcy:
I will test it against CEPH S3. 

I created this wiki page to document it https://github.com/gaul/s3proxy/wiki/Middleware-Tier-blobstore

src/main/java/org/gaul/s3proxy/TierBlobStore.java
```diff
@@ -0,0 +1,101 @@
...
+import org.jclouds.blobstore.util.ForwardingBlobStore;
+
+/**
+ * This class implements a middleware to set the storage tier when creating
+ * objects.  The class is configured via:
+ *
+ *   s3proxy.tier-blobstore = VALUE
+ *
+ * VALUE can be anything from org.jclouds.blobstore.domain.Tier, e.g.,
+ * STANRDARD, COOL, COLD, ARCHIVE.  org.jclouds.s3.domain.StorageClass
```
michaelcourcy, line 37: If I use COOL in the configuration 
```
s3proxy.tier-blobstore=COOL
```

I get this error message 
```
Exception in thread "main" java.lang.IllegalArgumentException: No enum constant org.jclouds.blobstore.domain.Tier.COOL
        at java.base/java.lang.Enum.valueOf(Enum.java:293)
        at org.jclouds.blobstore.domain.Tier.valueOf(Tier.java:24)
        at org.gaul.s3proxy.Main.parseMiddlewareProperties(Main.java:278)
        at org.gaul.s3proxy.Main.main(Main.java:125)
```

Look like we are limited by the enum values of [org.jclouds.blobstore.domain.Tier](https://jclouds.apache.org/reference/javadoc/2.1.x/org/jclouds/blobstore/domain/Tier.html) which are 

```
STANDARD
INFREQUENT
ARCHIVE
```

I would suggest to change this comment accordingly unless I pick up the wrong version of jcloud in the maven dependencies.

michaelcourcy (approved):
LGTM except the little change I suggest in the comments but not a blocker IMO.

src/main/java/org/gaul/s3proxy/TierBlobStore.java
michaelcourcy: Most likely that's what happened in this version there is COOL and COLD
https://github.com/apache/jclouds/blob/master/blobstore/src/main/java/org/jclouds/blobstore/domain/Tier.java
I did not figure out why when I built s3proxy but that does not matter please ignore my comment.

gaul: jclouds 2.6.0 only has a few Tier constants but 2.7.0-SNAPSHOT includes a wider variety.

gaul:
This now maps and logs S3 storage classes instead of jclouds Tiers as previously discussed.

src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -271,6 +272,19 @@ private static BlobStore parseMiddlewareProperties(BlobStore blobStore,
...
+                S3ProxyConstants.PROPERTY_STORAGE_CLASS_BLOBSTORE);
+        if (!Strings.isNullOrEmpty(storageClass)) {
+            System.err.println("Using storage class override backend");
+            var storageClassBlobStore =
+                    StorageClassBlobStore.newStorageClassBlobStore(
+                            blobStore, storageClass);
+            blobStore = storageClassBlobStore;
+            System.err.println("Requested tier: tier");
+            System.err.println("Overriden to: " +
+                    StorageClass.fromTier(storageClassBlobStore.getTier()));
```
gaul, line 285: This only works for S3-S3 mappings and is confusing for other storage backends.

michaelcourcy: I changed the wiki accordingly.

src/main/java/org/gaul/s3proxy/StorageClassBlobStore.java
```diff
@@ -0,0 +1,114 @@
...
+import org.jclouds.blobstore.domain.internal.BlobMetadataImpl;
+import org.jclouds.blobstore.options.PutOptions;
+import org.jclouds.blobstore.util.ForwardingBlobStore;
+import org.jclouds.s3.domain.ObjectMetadata.StorageClass;
+
+/**
+ * This class implements a middleware to set the storage tier when creating
+ * objects.  The class is configured via:
+ *
+ *   s3proxy.tier-blobstore = VALUE
```
michaelcourcy, line 33: Suggest : s3proxy.storage-class-blobstore = VALUE because of your recent changes

timuralp (approved):
Makes sense to me! Unfortunate that we're limited by the jclouds set of storage classes.

PR #287, opened 2018-11-14
Basic CORS support 

reimannf:
This PR brings the basic capability for `s3proxy` to respond to CORS preflight requests.

* S3ProxyHandler for HTTP method `OPTIONS`
* `OPTIONS` response include `Access-Control-Allow-Origin`, `Access-Control-Allow-Methods` and `Access-Control-Allow-Headers` Headers
* Rewrite `method` in signature generation for `OPTIONS` requests
* Configurability for allowed Origins, Methods and Headers for CORS Preflight requests
* Adding `Access-Control-Allow-Origin` header in CORS actual requests for GET, POST and PUT method based on provided `Origin` header
* `S3PROXY_CORS_ALLOW_ALL` allows every Origin and Header and Methods `GET, PUT, POST`

Allowed CORS settings with regex for Origins looks like:
```
docker run --publish 8888:80 \
   -e S3PROXY_... \
   -e S3PROXY_CORS_ALLOW_ORIGINS="https://.+\.example\.com https://.+\.example\.cloud http://localhost:.+" \
   -e S3PROXY_CORS_ALLOW_METHODS="GET PUT" \
   -e S3PROXY_CORS_ALLOW_HEADERS="Content-Type Accept" \
   -e JCLOUDS_... \
   --name s3proxy gaul/s3proxy:latest
```

This does not add support for S3 CORS bucket operations as such.

fixes #142

PR #668, opened 2024-07-30
Adds ETag to Access-Control-Expose-Headers headers

twick00:
I mentioned this in issue https://github.com/gaul/s3proxy/issues/667 but suffice to say that, while `ETag` is available as a header from many of the APIs in S3Proxy, the browser cannot access it since it's missing from the `Access-Control-Expose-Headers` list. To allow accessing this header from the browser, it needs to be added to the [Access-Control-Expose-Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers) header.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/CrossOriginResourceSharing.java
+++ b/src/main/java/org/gaul/s3proxy/CrossOriginResourceSharing.java
@@ -41,6 +41,7 @@ public final class CrossOriginResourceSharing {
     private static final String ALLOW_ANY_ORIGIN = "*";
     private static final String ALLOW_ANY_HEADER = "*";
     private static final String ALLOW_CREDENTIALS = "true";
+    private static final String EXPOSED_HEADERS = "ETag";
 
     private static final Logger logger = LoggerFactory.getLogger(
             CrossOriginResourceSharing.class);
@@ -101,6 +102,10 @@ public final class CrossOriginResourceSharing {
         logger.info("CORS allowed methods: {}", allowedMethods);
         logger.info("CORS allowed headers: {}", allowedHeaders);
         logger.info("CORS allow credentials: {}", allowCredentials);
+    }
+    
+    public String getExposedHeaders() {
+        return EXPOSED_HEADERS;
     }
 
     public String getAllowedMethods() {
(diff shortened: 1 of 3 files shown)


gaul:
@twick00 Please fix the stylistic issues.

@reimannf Does this seem correct to you?  Sorry to but you years after #287 but I don't have any background on this.

twick00:
I also created a [codesandbox example](https://codesandbox.io/p/sandbox/s3proxy-missing-etag-example-via-uppy-6q3d9f) that demonstrates the issue, but it has a few caveats:
1. Ensure the `endpoint` and `bucket` values are pointed to an S3 proxy configured with SSL
2. The S3Proxy should be running with `s3proxy.cors-allow-all=true` or properly configured SSL options

reimannf:
The extension to add `Access-Control-Expose-Headers` to the CORS response makes sense. Nevertheless I would make the list of Headers included in `Access-Control-Expose-Headers` configurable and would not hardcode `ETag` there.
@twick00 maybe you could look here https://github.com/gaul/s3proxy/blob/master/src/main/java/org/gaul/s3proxy/S3Proxy.java#L271-L272 as an example for allowed headers and pass it accordingly to `CrossOriginResourceSharing` extended constructor https://github.com/gaul/s3proxy/blob/master/src/main/java/org/gaul/s3proxy/CrossOriginResourceSharing.java#L62

twick00:
@reimannf, I had those changes ready to go locally so I created a PR (#674) instead of adding to this one, however, I'm not sure if it's the right solution. The fact that nobody has brought up the issue in the last 2 years shows that few people might care about customizing that `Access-Control-Expose-Headers` response so I decided to start with this PR since it seems like a much lighter solution. I didn't want to introduce a new configuration knob for people, IMO `Access-Control-Expose-Headers` should "just work".

Seems to me like the only 2 headers that most people care about are `etag` and `location`. [Heres an example](https://github.com/transloadit/uppy/blob/1ef39db22c585f30ef7a41dcd120f1972623d3a0/packages/@uppy/aws-s3/src/index.ts#L746-L774) of the lib that works with S3Proxy; notice that it only cares about those two headers I mentioned. For my specific case, I only care about `ETag`, but I could easily add `location` to the list of exposed headers.

If you really feel like you would prefer a new configuration knob then feel free to move over to #674 instead of this and I'll close this one.

reimannf:
Understand your arguments, but as you described for this specific lib, one would need already 2 headers.
I am not up to date regarding AWS S3 default CORS policy, but I think the `ETag` is not included there as `Access-Control-Expose-Headers`.
I would prefer an option, to configure the `Access-Control-Expose-Headers` as it would also allow custom headers to be included.
The final word has @gaul as I am not a maintainer, but my preference is #674

twick00:
I'll close this for now and we can migrate to #674

gaul:
Thank you @reimannf for your feedback!

PR #674, opened 2024-08-05
Adds feature allowing Access-Control-Expose-Headers configuration

twick00:
Copy/paste of #668 description:

> I mentioned this in issue https://github.com/gaul/s3proxy/issues/667 but suffice to say that, while ETag is available as a header from many of the APIs in S3Proxy, the browser cannot access it since it's missing from the Access-Control-Expose-Headers list. To allow accessing this header from the browser, it needs to be added to the [Access-Control-Expose-Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers) header.

This PR introduces a new config variable; `s3proxy.cors-exposed-headers` (or `S3PROXY_CORS_EXPOSED_HEADERS`) that allows configuring the values returned with `Access-Control-Expose-Headers` for CORS.

### DIFF
--- a/src/test/java/org/gaul/s3proxy/CrossOriginResourceSharingRuleTest.java
+++ b/src/test/java/org/gaul/s3proxy/CrossOriginResourceSharingRuleTest.java
@@ -18,6 +18,8 @@ package org.gaul.s3proxy;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.util.List;
+
 import com.google.common.collect.Lists;
 
 import org.junit.Before;
@@ -39,7 +41,7 @@ public final class CrossOriginResourceSharingRuleTest {
                         "https://example\\.cloud"),
                 Lists.newArrayList("GET", "PUT"),
                 Lists.newArrayList("Accept", "Content-Type"),
-                Lists.newArrayList(),
+                List.of(),
                 "true");
         // CORS disabled
         corsOff = new CrossOriginResourceSharing(null, null, null, null, null);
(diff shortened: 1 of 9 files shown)


src/test/java/org/gaul/s3proxy/CrossOriginResourceSharingRuleTest.java
```diff
@@ -39,9 +39,10 @@ public void setUp() throws Exception {
                         "https://example\\.cloud"),
                 Lists.newArrayList("GET", "PUT"),
                 Lists.newArrayList("Accept", "Content-Type"),
+                Lists.newArrayList(),
                 "true");
         // CORS disabled
-        corsOff = new CrossOriginResourceSharing(null, null, null, null);
+        corsOff = new CrossOriginResourceSharing(null, null, null, null, null);
```
gaul, line 45: Is there any kind of test that can exercise how this should work when it is configured/not configured?  s3-tests might have some functionality so we might just need to enable tests there.  Currently CI disables all the cors tests:

```
src/test/resources/run-s3-tests.sh:./virtualenv/bin/nosetests -a '!fails_on_s3proxy,!appendobject,!bucket-policy,!cors,!encryption,!fails_strict_rfc2616,!lifecycle,!object-lock,!policy,!policy_status,!s3select,!s3website,!tagging,!user-policy,!versioning'
```

We can decouple testing from this PR if you think it makes sense.

twick00: I don't see much value in including a test to check the exposed headers from this file since it would just be a simple passthrough-existance check . I can, however, investigate adding some better tests for `Access-Control-Expose-Headers` in a follow-on PR if thats alright.

I will also update the docs related to the new `s3proxy.cors-exposed-headers` (`S3PROXY_CORS_EXPOSED_HEADERS`) properties in that PR if that works for you.

gaul: Yeah let's add the properties at least.  I opened #675 to track possible testing but let's not block this PR on that.

gaul: Oh I see you already added the Docker environment variable.

gaul:
Thank you for your contribution @twick00!

PR #678, opened 2024-08-13
README:  Adobe s3mock

sullis:
Findify s3mock is archived.

Adobe s3mock is an active project.


gaul:
Thanks for your contribution @sullis!  I am happy to coordinate with the new fork to reduce duplication between the two projects if you have any ideas.

PR #679, opened 2024-08-28
Allow x-amz-user-agent header

losfair:
The Rust S3 SDK `aws-sdk-s3` (https://github.com/awslabs/aws-sdk-rust) sends the `x-amz-user-agent` header, but `s3proxy` does not know about it. This causes s3proxy to reject all requests from `aws-sdk-s3`.

This patch adds `x-amz-user-agent` to the `SUPPORTED_X_AMZ_HEADERS` list.


gaul:
Thank you for your contribution @losfair!

PR #680, opened 2024-08-30
allow invalid auth header for presigned urls

RunFMe:
I have a case where I upload files with presigned links and my requests already contain Authentication header for another application. The app is tested against regular s3, which ignores the header and parses info from presigned link. s3proxy only parses parameters from url if the Authentication header is not present. 
I suggest we change the behavior to mimick s3 better: we will will try to construct S3AuthorizationHeader from Auth header and if it fails or header is not present try to parse url parameters.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -366,13 +366,27 @@ public class S3ProxyHandler {
 
         BlobStore blobStore;
         String requestIdentity = null;
-        String headerAuthorization = request.getHeader(
-                HttpHeaders.AUTHORIZATION);
         S3AuthorizationHeader authHeader = null;
         boolean presignedUrl = false;
+        boolean headerAuth = true;
 
         if (!anonymousIdentity) {
+            //try login with auth header
+            String headerAuthorization = request.getHeader(HttpHeaders.AUTHORIZATION);
+            
             if (Strings.isNullOrEmpty(headerAuthorization)) {
+                headerAuth = false;
+            } else {
+                try {
+                    authHeader = new S3AuthorizationHeader(headerAuthorization);
+                    //whether v2 or v4 (normal header and query)
+                } catch (IllegalArgumentException iae) {
+                    headerAuth = false;
+                }
+            }
+
+            //try login as pre signed url if header failed
+            if (!headerAuth) {
                 String algorithm = request.getParameter("X-Amz-Algorithm");
                 if (algorithm == null) { //v2 query
                     String identity = request.getParameter("AWSAccessKeyId");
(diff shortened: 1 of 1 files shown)


gaul:
Please address the Checkstyle and s3-tests failures:

```
======================================================================
FAIL: s3tests.functional.test_headers.test_object_create_bad_authorization_invalid_aws2
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gaul/work/s3proxy/s3-tests/virtualenv/lib/python3.6/site-packages/nose/case.py", line 170, in runTest
    self.test(*self.arg)
  File "/home/gaul/work/s3proxy/s3-tests/s3tests/functional/test_headers.py", line 414, in test_object_create_bad_authorization_invalid_aws2
    eq(e.status, 400)
AssertionError: 403 != 400

======================================================================
FAIL: s3tests.functional.test_headers.test_bucket_create_bad_authorization_invalid_aws2
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gaul/work/s3proxy/s3-tests/virtualenv/lib/python3.6/site-packages/nose/case.py", line 170, in runTest
    self.test(*self.arg)
  File "/home/gaul/work/s3proxy/s3-tests/s3tests/functional/test_headers.py", line 443, in test_bucket_create_bad_authorization_invalid_aws2
    eq(e.status, 400)
AssertionError: 403 != 400
```

gaul:
@RunFMe Can you fix the test failures as I suggested?  I plan to run a release this weekend.

gaul:
@RunFMe could you finish this PR?

gaul:
@RunFMe could you finish this PR?

gaul:
Thanks for this, and apologies it sat so long.

The problem is real and still reproduces on master. A presigned GET succeeds
on its own; the same URL with `Authorization: Bearer some-token` returns
400 InvalidArgument. Your reading of the cause was right, and the scenario --
one HTTP stack talking to this proxy and to something else that wants its own
Authorization header -- is not one I would have gone looking for.

The patch stops one layer short, though. `doHandle` is only the first of two
places that read the header. `AwsSignature.createCanonicalRequest` reads it
again and uses it for two more decisions: which headers the signature covers,
and whether to hash the request body or treat it as `UNSIGNED-PAYLOAD` the way
a presigned request does. With this branch applied the same request moves from
400 InvalidArgument to 403 AccessDenied -- still refused, just under a
different code, which is an easy thing to mistake for a fix on review.

The complete version is on master as 05b8b93: the presigned decision is made
once and passed down, so both places agree on where the signature is coming
from. One
deliberate narrowing over what is proposed here -- the fallback triggers on the
header naming a scheme that is not ours, not on the parse failing. A header
that says `AWS4-HMAC-SHA256` and then does not parse stays a 400, because
falling back there would let a caller replace a header they cannot sign with
query parameters they can.

Tests cover a bearer token, a basic credential, a malformed AWS header, a
presigned URL on its own, and a foreign header with no presigning at all -- the
last so that an unrelated header can never turn an unsigned request into a
signed one. s3-tests reports 264 passed with no failures.

Two notes on the branch as it stands, both from age rather than anything you
did: it no longer compiles against master, since NullAway landed after you
opened this and cannot prove `authHeader` is non-null through the new control
flow; and there is a trailing-whitespace line in the added block.

Credit to you for the report and for the diagnosis -- both were right, and the
fix is the one you pointed at, carried one layer further.

PR #696, opened 2024-10-02
Add azureblob-sdk provider backed by Azure SDK

### LINKED ISSUE #606: Add an Azure storage backend that uses the Microsoft SDK
S3Proxy currently uses Apache jclouds for its Azure Blob integration.  This uses the native REST API and has some advantages in terms of similar configuration and logging.  However the [Microsoft SDK](https://learn.microsoft.com/en-us/java/api/overview/azure/storage?view=azure-java-stable) has many more features including authentication that are popular user requests, e.g., #536.  This might not be a lot of work to just create a `BlobStore` implementation that calls into the SDK.

gaul:
This may replace the jclouds azureblob provider.  The implementation lacks some error codes and options.  Fixes #606.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Maven Central hosts S3Proxy artifacts and the wiki has
 * atmos
 * aws-s3 (Amazon-only)
 * azureblob
+* azureblob-sdk (newer but incomplete)
 * b2
 * filesystem (on-disk storage)
 * google-cloud-storage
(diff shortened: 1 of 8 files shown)


gaul:
While this is incomplete, I would appreciate if some of the Azure issue reporters could test this to give some early feedback.  Using the SDK seems like a better path forward than the existing jclouds approach.  Using the SDK should allow more authorization options and it would be great if people could help me understand them so I can add configuration knobs.  Some downsides of this PR are that there may be more conflicts between dependencies, e.g., Guava, but this seems tolerable.

PR #699, opened 2024-10-07
Support SSL setup via SSLContext in S3Proxy.Builder

bersti:
Simplify S3Proxy usage in Java when an SSLContext instance is already at hand and creating keystore files is inconvenient.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3Proxy.java
+++ b/src/main/java/org/gaul/s3proxy/S3Proxy.java
@@ -25,6 +25,8 @@ import java.util.Collection;
 import java.util.Objects;
 import java.util.Properties;
 
+import javax.net.ssl.SSLContext;
+
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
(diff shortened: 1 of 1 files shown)


src/main/java/org/gaul/s3proxy/S3Proxy.java
```diff
@@ -107,8 +111,12 @@ public final class S3Proxy {
...
             SslContextFactory.Server sslContextFactory =
                 new SslContextFactory.Server();
-            sslContextFactory.setKeyStorePath(builder.keyStorePath);
-            sslContextFactory.setKeyStorePassword(builder.keyStorePassword);
+            if (builder.sslContext != null) {
+                sslContextFactory.setSslContext(builder.sslContext);
+            } else {
+                sslContextFactory.setKeyStorePath(builder.keyStorePath);
+                sslContextFactory.setKeyStorePassword(builder.keyStorePassword);
+            }
```
gaul, line 119: Do you think it is better to require the caller to create a `SslContextFactory.Server` then pass it to the `Builder`?

bersti: I also considered that option, but I  thought it would be better to avoid exposing Jetty specifics/internals (like SslContextFactory.Server) to the outside world, and minimize API changes if S3proxy was ever to swap Jetty for something else.

gaul: Makes sense.

gaul:
Thank you for your contribution @bersti!

PR #702, opened 2024-10-08
Add middleware to replace user metadata

### LINKED ISSUE #466: Illegal characters in Blob metadata not handled gracefully
Hi. I'm investigating S3Proxy for suitability in a use case where I want a third-party application to use Azure Blob Storage via the S3 API. The third-party application is not open for modification, so I have to make do with whatever it is throwing at the API.

I'm running into an issue where the application is attempting to store metadata with a blob that contain illegal characters (i.e. a dash '-'). S3Proxy throws a 500 Internal Server error when this happens.

Encoding (and decoding on reads) may not be feasible. But I can imagine an opt-in feature that, if enabled, will store metadata in the Blob Index Tags instead of in Metadata. See https://learn.microsoft.com/en-us/azure/storage/blobs/storage-manage-find-blobs?tabs=azure-portal#choosing-between-metadata-and-blob-index-tags for details.

From the looks of it, the current behavior is blocking my use of S3Proxy, so I would be very appreciative if this improvement would be added.

gaul:
This is useful when keys and values must conform to some subset of values, e.g., Azure's C# identifiers.  Fixes #466.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ S3Proxy can modify its behavior based on middlewares:
 * [regex rename blobs](https://github.com/gaul/s3proxy/wiki/Middleware-regex)
 * [sharded backend containers](https://github.com/gaul/s3proxy/wiki/Middleware-sharded-backend)
 * [storage class override](https://github.com/gaul/s3proxy/wiki/Middleware-storage-class-override)
+* user metadata replacer
 
 ## SSL Support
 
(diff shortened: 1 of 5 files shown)


gaul:
Does single-character replacement suffice?  Or would something more involved like regexes address more use cases?

atsai1220:
Could be another option `s3proxy.user-metadata-replacer-blobstore-regex=true/false`. Thoughts?

gaul:
But what is the use case for more complicated replacements?

atsai1220:
I'm happy with a straightforward string match and replacement.

PR #705, opened 2024-10-10
Allow setting service-path config when running via Docker

ferjm:
This update allows the `s3proxy.service-path` configuration introduced in https://github.com/gaul/s3proxy/pull/194 to be set when running S3Proxy via Docker


gaul:
Thank you for your contribution @ferjm!

PR #706, opened 2024-10-11
Set outputLength correctly on open ended range requests to encryptedBlobStore

ptemarvelde:
This resolves https://github.com/gaul/s3proxy/issues/698.

Open ended ranged requests to encrypted files resulted in incorrect Content-Length headers in the response because of a bug in the Decryption class.

In EncryptedBloBStore the length is set to `-1` by default. On open-ended ranged GET requests this value is passed to the `Decryption` constructor, which in turn only sets `outputLength` if an `offset` is given without a `length`, but `-1` is used to represent no length given. After this change the `outputLength` is set correctly in this constructor.

https://github.com/gaul/s3proxy/blob/master/src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java#L375

@gaul The tests pass (locally) but I have not spend enough time with this codebase to reason about whether this change will break anything else (which is currently untested). Is this a reasonable solution for https://github.com/gaul/s3proxy/issues/698?

### DIFF
--- a/src/test/java/org/gaul/s3proxy/EncryptedBlobStoreLiveTest.java
+++ b/src/test/java/org/gaul/s3proxy/EncryptedBlobStoreLiveTest.java
@@ -159,12 +159,16 @@ public final class EncryptedBlobStoreLiveTest extends S3ClientLiveTest {
         options.range(0, 19);
         object = this.getApi().getObject(containerName, blobName, options);
 
-        InputStreamReader r =
-            new InputStreamReader(object.getPayload().openStream());
+        InputStreamReader r = new InputStreamReader(
+                object.getPayload().openStream());
         BufferedReader reader = new BufferedReader(r);
-        String partialContent = reader.lines().collect(Collectors.joining());
-        assertThat(partialContent).isEqualTo(content.substring(0, 20));
-        assertThat((long) partialContent.length()).isEqualTo(object.getMetadata().getContentMetadata().getContentLength());
+        String partialContent = reader.lines()
+                .collect(Collectors.joining());
+
+        assertThat(partialContent).isEqualTo(
+                content.substring(0, 20));
+        assertThat((long) partialContent.length()).isEqualTo(
+                object.getMetadata().getContentMetadata().getContentLength());
 
         // open ended range request
         options = new GetOptions();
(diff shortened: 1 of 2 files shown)


gaul:
Can you add a unit test that demonstrates this problem?  I wonder how does encryption work for range requests?  I would imagine that S3Proxy needs to decrypt either the entire object or at least whatever the encryption block size is.  CC @FlorinPeter who wrote the original code.

ptemarvelde:
Partial decryption is already extensively tested in `EncryptedBlobStoreTest`, see [`testReadPartial`](https://github.com/gaul/s3proxy/blob/33b3427256a5f4ae46d34177337a5e216d0bbd41/src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java#L564), [`testMultiPartReadPartial`](https://github.com/gaul/s3proxy/blob/33b3427256a5f4ae46d34177337a5e216d0bbd41/src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java#L670) and [`testMultiPartReadTail`](https://github.com/gaul/s3proxy/blob/33b3427256a5f4ae46d34177337a5e216d0bbd41/src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java#L721). Both `testMultiPartRead*` tests test the situation where no explicit end byte is given. 

As confirmed by these tests, the content s3proxy returns for encrypted blobs for any ranged request is correct. Just the ContentLength response header is incorrect in the described case, due to the bug in the `Encryption` class. I have extended one of the unittests to verify this behavior and show the content length is set correctly now.

src/test/java/org/gaul/s3proxy/EncryptedBlobStoreLiveTest.java
```diff
@@ -164,6 +164,18 @@ public void testPartialContent() throws InterruptedException, IOException {
...
+        // open ended range request
+        options = new GetOptions();
+        options.startAt(10);
+        object = this.getApi().getObject(containerName, blobName, options);
+
+        r = new InputStreamReader(object.getPayload().openStream());
+        reader = new BufferedReader(r);
+        partialContent = reader.lines().collect(Collectors.joining());
+        assertThat(partialContent).isEqualTo(content.substring(10));
+        assertThat((long) partialContent.length()).isEqualTo(object.getMetadata().getContentMetadata().getContentLength());
```
gaul, line 178: Can you address the overlong lines Checkstyle violation?

gaul (changes requested):
Thanks for the explanation!  Please fit the nits so I can merge this.

gaul:
Thank you for your contribution and explanation @ptemarvelde!  I didn't write this middleware and don't have a good understanding how it works.

PR #707, opened 2024-10-12
Add NIO.2 BlobStore

### LINKED ISSUE #697: Add a nio2 filesystem backend
Currently S3Proxy uses jclouds for the transient (in-memory) and filesystem (on-disk) providers for local storage.  Given jclouds declining activity, it would be prudent to bring more of its functionality into S3Proxy itself which would simplify debugging and releasing.  Implementing a nio2 filesystem backend would allow not only local on-disk storage, but also memory via [jimfs](https://github.com/google/jimfs), and potentially crazier things like [Hadoop](https://github.com/damiencarol/jsr203-hadoop).  References #606.

gaul:
This will enable multiple backends, e.g., jimfs (in-memory), filesystem, and possibly stranger things like Hadoop.  Currently only configured to use jimfs.  Fixes #697.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -75,6 +75,11 @@ jobs:
         run: |
           mvn test
 
+      - name: Maven Test with transient-nio2
+        run: |
+          # TODO: run other test classes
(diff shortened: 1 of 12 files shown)


gaul:
This still lacks list marker support and blob access control but it is good enough to pass the s3fs integration tests so let's merge it.

PR #719, opened 2024-11-11
Run s3-tests against Azurite in CI

### LINKED ISSUE #700: Use Azurerite to for Azure CI testing
Microsoft offers [Azurite](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite) to allow local testing of Azure blobstore.  This could be used for S3Proxy Azure CI testing.  Initially just use the unit tests but s3-tests should be possible as well.

gaul:
Fixes #700.


gaul:
A few tests will fail until (Azure/Azurite#2462 or Azure/Azurite#2486) and Azure/Azurite#2487 merge.

Issue #536, opened 2023-08-01
Azure managed Identity


akshay-sawant-bh:
Not able to connect to azure managed identity using client id getting issue, is there a set of steps to achieve this, has anyone tried this

alifirat:
Hi what his the Azure auth method you have used ? It's a jclouds setting to set.

akshay-sawant-bh:
We are trying to use Azure managed Identity using the clientId to connect to Azure but its not working

gaul:
@akshay-sawant-bh I implemented a new azureblob-sdk provider in #606.  This should support alternative authentication mechanisms but needs some additional configuration to use these instead of the access keys.  Could you look into this?

PR #728, opened 2024-11-22
Allow to use Azure managed identities for azureblob-sdk backend

### LINKED ISSUE #352: Cannot use S3Proxy with Azure Pod Identities
We are trying to deploy S3Proxy in a Kubernetes cluster on AKS in order to translate AWS S3-compatible requests performed by our own software to Azure Blob Storage requests. Even though passing an Azure storage account token to the `JCLOUDS_CREDENTIAL` variable works, we are having trouble authenticating using Azure Pod Identity credentials.

In a nutshell, Pod Identities allow pods running in an AKS cluster to retrieve a JWT from the Azure Instance Metadata Service and use it to assume a managed identity when accessing Azure services. We want to use this feature because it allows assigning fine-grained permissions to services running in the AKS cluster, and is considered best practice for applications running on AKS. See more information here:

https://docs.microsoft.com/en-us/azure/aks/use-azure-ad-pod-identity
https://docs.microsoft.com/en-us/azure/aks/operator-best-practices-identity#use-pod-managed-identities

When trying to use this feature in combination with S3Proxy, we faced the following two problems:

1. S3Proxy does not automatically retrieve and refresh Pod Identity credentials from the metadata service. This is a problem because these credentials are short-lived, and are expected to be refreshed regularly by the application.

2. When passing a Pod Identity token retrieved from the metadata service to S3Proxy via the `JCLOUDS_CREDENTIAL` variable, requests towards S3Proxy fail with the following error:

   ```HTML
   <html>
   <head>
   <meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
   <title>Error 400 </title>
   </head>
   <body>
   <h2>HTTP ERROR: 400</h2>
   <p>Problem accessing /. Reason:
   <pre>    com.google.common.io.BaseEncoding$DecodingException: Unrecognized character: .</pre></p>
   <hr /><i><small>Powered by Jetty://</small></i>
   </body>
   </html>
   ```

   What we think is happening is that jclouds is trying to base64 decode the Pod Identity token in order to sign the request. However, since this token is a JWT, decoding fails with the above error once it encounters the first dot.

Is there any way to overcome these problems and use S3Proxy with Pod Identities right now? Is there any plan to add support for them in the future?

hermandavid:
This PR closes #352, #536.

Current behaviour is unaffected. If backend provider is `azureblob-sdk`, and both `identity`, and `credentials` are empty strings, authentication fallbacks to [DefaultAzureCredential](https://learn.microsoft.com/en-us/java/api/com.azure.identity.defaultazurecredential?view=azure-java-stable) and tries to login with multiple approaches, including managed identities.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
+++ b/src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
@@ -33,6 +33,7 @@ import javax.annotation.Nullable;
 
 import com.azure.core.credential.AzureNamedKeyCredential;
 import com.azure.core.http.rest.PagedResponse;
+import com.azure.identity.DefaultAzureCredentialBuilder;
 import com.azure.storage.blob.BlobServiceClient;
 import com.azure.storage.blob.BlobServiceClientBuilder;
 import com.azure.storage.blob.models.AccessTier;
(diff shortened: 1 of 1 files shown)


gaul:
This looks reasonable -- did it work in your testing?

hermandavid:
Yes, I have tested it in our AKS environment with success

```
kubectl create namespace s3proxytest

kubectl -n s3proxytest create deployment s3proxy --image=xxx/s3proxy:master

kubectl -n s3proxytest set env deployment/s3proxy \
	JCLOUDS_PROVIDER=azureblob-sdk \
	JCLOUDS_IDENTITY="" \
	JCLOUDS_CREDENTIAL="" \
	JCLOUDS_ENDPOINT="https://xxx.blob.core.windows.net" \
	S3PROXY_AUTHORIZATION="none"

kubectl -n s3proxytest exec -it deployments/s3proxy -- curl http://localhost

kubectl delete namespaces s3proxytest
```

gaul:
Thank you for your contribution @hermandavid!

gaul:
@hermandavid Could you add another section to https://github.com/gaul/s3proxy/wiki/Storage-backend-examples explaining how to use azureblob-sdk and the multiple credential modes?

hermandavid:
@gaul Sure, I have updated the wiki.

Issue #711, opened 2024-10-30
Use fake-gcs-server for CI


gaul:
Similar to #700, using [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) would allow better testing of GCS.

gaul:
This will be tough in the short-term since the jclouds GCS provider calls out to OAuth unconditionally and does not appear to have an anonymous mode.  But the referenced commit is useful for local testing.

gaul:
Fixed by 38b738754b0afcc53789498a5a22acddb5fc0362.

PR #730, opened 2024-11-25
Experiment with fake-gcs-server

gaul:
References #711.


gaul:
This won't immediately work since jclouds reaches out to google.com to do oauth.  I guess I could add anonymous access to jclouds but it's probably a better long-term solution to write a provider using the Google Java SDK.

gaul:
Done in 38b738754b0afcc53789498a5a22acddb5fc0362.

PR #744, opened 2024-12-24
Run s3-tests against minio

gaul:
Also clean up some incorrect tagging.


gaul:
Several tests will fail until S3Proxy can upgrade to Apache jclouds 2.6.1.

PR #749, opened 2024-12-27
chore: support IRSA for aws s3 provider

Anhui-tqhuang:


src/main/java/org/gaul/s3proxy/Main.java
```diff
@@ -386,6 +388,20 @@ private static BlobStore createBlobStore(Properties properties,
...
 
+        if ((identity.isEmpty() || credential.isEmpty()) && provider.equals("aws-s3")) {
+            DefaultAWSCredentialsProviderChain creds = DefaultAWSCredentialsProviderChain.getInstance();
+            Supplier<Credentials> credentialsSupplier = new Supplier<Credentials>() {
+                @Override
+                public Credentials get() {
+                    return new Credentials(creds.getCredentials().getAWSAccessKeyId(),
+                            creds.getCredentials().getAWSSecretKey());
+                }
+            };
```
gaul, line 399: Could you use the Apache jclouds STS supplier for consistency:

https://stackoverflow.com/questions/23520216/using-aws-s3-via-jclouds-how-to-assume-role

pom.xml
```diff
@@ -545,5 +550,11 @@
       <artifactId>javax.annotation-api</artifactId>
       <version>1.3.2</version>
     </dependency>
+      <dependency>
+          <groupId>org.jetbrains</groupId>
+          <artifactId>annotations</artifactId>
+          <version>RELEASE</version>
+          <scope>compile</scope>
+      </dependency>
```
gaul, line 558: What does this do?

gaul (commented):
Can you add an example to the wiki like the Azure one: https://github.com/gaul/s3proxy/wiki/Storage-backend-examples#azure-blob-sdk---managed-identity ?

gaul (changes requested):
Please alphabetize imports as Checkstyle suggests.

src/main/java/org/gaul/s3proxy/Main.java
Anhui-tqhuang: hey, i am still testing this change, btw, the sts api only supports assume role not assume role with web identity 🤔

Anhui-tqhuang:
@gaul hey, page updated https://github.com/gaul/s3proxy/wiki/Storage-backend-examples

and my change tested
```console
$ aws s3 ls xxx --endpoint http://0.0.0.0:4449
                           PRE customer-pg-backups/
...
```

pom.xml
```diff
@@ -382,6 +382,11 @@
   </repositories>
 
   <dependencies>
+    <dependency>
+      <groupId>commons-logging</groupId>
+      <artifactId>commons-logging</artifactId>
+      <version>1.2</version>
```
gaul, line 388: What does this do?

src/main/java/org/gaul/s3proxy/Main.java
gaul: Could you expand on your comment?  Do you mean that the AWS STS library only supports assume role?  Or do you mean the jclouds code only allows it?  If the former, it would be better to remove the aws-java-sdk-sts dependency.  But if the latter, I am willing to take this if you can explain more clearly what this means.

gaul (changes requested):

pom.xml
Anhui-tqhuang: without this, aws sdk will not be able to work with this error due to this exclude
```
    <dependency>
      <groupId>com.amazonaws</groupId>
      <artifactId>aws-java-sdk-s3</artifactId>
      <version>1.12.261</version>
      <scope>test</scope>
      <exclusions>
        <exclusion>
          <groupId>commons-logging</groupId>
          <artifactId>commons-logging</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
```

src/main/java/org/gaul/s3proxy/Main.java
Anhui-tqhuang: sure

pom.xml
gaul: I don't remember the full context for this from 8 years ago but e569b190f2b9996bd89fae8f2b7ed888fb081c3b suggests that I configured the AWS SDK to use the sl4fj logging like the rest of S3Proxy/jclouds.  Can you investigate this further?

src/main/java/org/gaul/s3proxy/Main.java
Anhui-tqhuang: assume role is:
- you have your own credentials (access key id and access secret key), then you have the role arn that you want to assume
- assume role with web identity is, you have only one role arn and one web identity token

jclouds doesn't have a way to let you load you default aws credentials including web identity token

pom.xml
Anhui-tqhuang: i tried to remove that dependency and i got this error
```
Caused by: java.lang.NoClassDefFoundError: Could not initialize class com.amazonaws.auth.DefaultAWSCredentialsProviderChain
	at org.gaul.s3proxy.Main$2.get(Main.java:399)
	at org.gaul.s3proxy.Main$2.get(Main.java:396)
	at org.jclouds.rest.internal.RestAnnotationProcessor.apply(RestAnnotationProcessor.java:241)
	at org.jclouds.rest.internal.RestAnnotationProcessor.apply(RestAnnotationProcessor.java:137)
	at org.jclouds.rest.internal.InvokeHttpMethod.toCommand(InvokeHttpMethod.java:189)
	at org.jclouds.rest.internal.InvokeHttpMethod.invoke(InvokeHttpMethod.java:85)
	at org.jclouds.rest.internal.InvokeHttpMethod.apply(InvokeHttpMethod.java:74)
	at org.jclouds.rest.internal.InvokeHttpMethod.apply(InvokeHttpMethod.java:45)
	at org.jclouds.rest.internal.DelegatesToInvocationFunction.handle(DelegatesToInvocationFunction.java:156)
	at org.jclouds.rest.internal.DelegatesToInvocationFunction.invoke(DelegatesToInvocationFunction.java:123)
	at jdk.proxy2/jdk.proxy2.$Proxy65.getBucketLocation(Unknown Source)
	at org.jclouds.s3.config.S3HttpApiModule$3.load(S3HttpApiModule.java:129)
	at org.jclouds.s3.config.S3HttpApiModule$3.load(S3HttpApiModule.java:125)
	at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3574)
	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2316)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2189)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2079)
```

Anhui-tqhuang: i figured out that the program could only work when this dependency is added 😢

Anhui-tqhuang: any hints? 🤔 as i am not familiar with java really 😢

Anhui-tqhuang: i saw these imports
```java
import org.apache.commons.logging.Log;
import org.apache.commons.logging.LogFactory;
```

in the class definition
```java
public class AWSCredentialsProviderChain implements AWSCredentialsProvider {
....
}
```
I don't know if it is possible to replace the logging that the package has imported 🤔

src/main/java/org/gaul/s3proxy/Main.java
Anhui-tqhuang: btw, default aws credential provider chan here could help the program to load default credentials from many kind of aws sdk compatible env like
- aws credentials: AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
- iam role service account: AWS_ROLE_ARN, AWS_WEB_IDENTITY_TOKEN_FILE
- etc ...

pom.xml
gaul: Could you remove the `<scope>test</scope>` from `<artifactId>jcl-over-slf4j</artifactId>`?  This should address the error.

Anhui-tqhuang: @gaul yep, that works!, thanks! i removed the common logging dependency

gaul:
Thank you for your contribution @Anhui-tqhuang!

Anhui-tqhuang:
@gaul thx a lot!

Anhui-tqhuang:
@gaul and thanks for fixing https://github.com/gaul/s3proxy/commit/3ff077f5504ba13f982dea6e98644d6dff2128ee

i just noticed that and spike how to fix it 🤣

PR #761, opened 2025-01-18
Use ubuntu-24.04-arm GitHub runner

gaul:


gaul:
This seems to reduce runTests times from 10 minutes on x86-64 to 8 minutes on arm64.

PR #763, opened 2025-01-21
Remove explicit dependency on Apache commons-io

Aazme:
commons-fileupload2-javax exposes this as an implicit dependency but
S3Proxy can replace its uses with Guava and modern Java.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -293,6 +293,12 @@ public class S3ProxyHandler {
         String uri = request.getRequestURI();
         String originalUri = request.getRequestURI();
 
+        // Check for the /version endpoint
+        if ("/healthz".equals(uri) && "GET".equalsIgnoreCase(method)) {
+            handleVersionRequest(response);
+            return;
+        }
+
         if (!this.servicePath.isEmpty()) {
             if (uri.length() > this.servicePath.length()) {
                 uri = uri.substring(this.servicePath.length());
@@ -2028,6 +2034,18 @@ public class S3ProxyHandler {
         addCorsResponseHeader(request, response);
 
         response.addHeader(HttpHeaders.ETAG, maybeQuoteETag(eTag));
+    }
+    private void handleVersionRequest(HttpServletResponse response) throws IOException {
+        response.setStatus(HttpServletResponse.SC_OK);
+        response.setContentType("application/json");
+        response.setCharacterEncoding("UTF-8");
+    
+        String versionInfo = "{ \"status\": \"OK\"}";
+    
+        try (PrintWriter writer = response.getWriter()) {
+            writer.write(versionInfo);
+            writer.flush();
+        }
     }
 
     private void handlePostBlob(HttpServletRequest request,


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2029,6 +2035,18 @@ private void handlePutBlob(HttpServletRequest request,
 
         response.addHeader(HttpHeaders.ETAG, maybeQuoteETag(eTag));
     }
+    private void handleVersionRequest(HttpServletResponse response) throws IOException {
```
gaul, line 2038: Should this be called `handleStatuszRequest`?

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2029,6 +2035,18 @@ private void handlePutBlob(HttpServletRequest request,
...
+        response.setStatus(HttpServletResponse.SC_OK);
+        response.setContentType("application/json");
+        response.setCharacterEncoding("UTF-8");
+    
+        String versionInfo = "{ \"status\": \"OK\"}";
```
gaul, line 2043: Extra space between `{` and `\"`.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -2029,6 +2035,18 @@ private void handlePutBlob(HttpServletRequest request,
 
         response.addHeader(HttpHeaders.ETAG, maybeQuoteETag(eTag));
     }
```
gaul, line 2037: Add newline before new method.

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -293,6 +293,12 @@ public final void doHandle(HttpServletRequest baseRequest,
         String uri = request.getRequestURI();
         String originalUri = request.getRequestURI();
 
+        // Check for the /version endpoint
```
gaul, line 296: Remove useless comment.

gaul (changes requested):
Please address Checkstyle warnings:

```
Warning:  src/main/java/org/gaul/s3proxy/S3ProxyHandler.java:[2042] (regexp) RegexpSingleline: Line has trailing spaces.
Warning:  src/main/java/org/gaul/s3proxy/S3ProxyHandler.java:[2044] (regexp) RegexpSingleline: Line has trailing spaces.
```

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
gaul: Maybe this should return the current time, the time s3proxy launched, and the git hash?  Are there any well-known statusz formats this could follow?

gaul (changes requested):
Can you exercise this in some kind of test, e.g., `AwsSdkAnonymousTest`?

gaul:
I'm preparing a new release so please try to address my comments before the weekend.

Aazme:
I'll work on those during the coming days, had a busy weeks

musabshak:
> This PR introduces a new /healthz endpoint in S3Proxy to provide a basic health check mechanism.

For the "has my app started yet" aspect of "basic health check mechanism" in k8s, you can also do:

```
        livenessProbe:
          tcpSocket:
            port: 80
```

The docker image exposes port 80 and the Java app/server is listening on port 80. Until the Java server is initialized, the TCP connection on port 80 won't be able to be established. 

The addition of an explicit `/healthz` HTTP endpoint (that is un-authed) in the PR is useful though because it lays the foundation for doing a more meaningful "health check". For example, if the backend provider is GCS, the health check should ping the GCS endpoint to make sure it's reachable.

pom.xml
```diff
@@ -168,6 +168,26 @@
           </archive>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>pl.project13.maven</groupId>
+        <artifactId>git-commit-id-plugin</artifactId>
+        <version>5.0.0</version>
```
gaul: How did you test this locally?  The latest version of git-commit-id-plugin is 4.9.10: https://mvnrepository.com/artifact/pl.project13.maven/git-commit-id-plugin while the latest version of git-commit-id-maven-plugin is 9.0.2: git-commit-id-maven-plugin

gaul (changes requested):

gaul:
@Aazme Can you finish this PR?

gaul:
Thank you for your contribution @Aazme!

PR #764, opened 2025-01-21
feat: Support json logging

ryanfaircloth:


ryanfaircloth:
Most log systems actually will prefer it Splunk, CrowdStrike logscale and others. In particular it helps with multiline logoutput for container deployments because its difficult to assign config to file readers to deal with start and end events by using json the agents and destination systems at least don't have to get the start/end right

gaul:
Thanks for the explanation!  Please let me know if there is anything S3Proxy could do to fit better into your environment.

PR #815, opened 2025-05-23
fix: service path expected uri for signature to match.

re1709:
Using authentication with a service path (e.g. under kubernetes deployment) does not work, as the expected uri is incorrect (contains the service path twice).

### DIFF
--- a/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
+++ b/src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
@@ -630,8 +630,8 @@ public class S3ProxyHandler {
                         is = new ByteArrayInputStream(payload);
                     }
 
-                    String uriForSigning = presignedUrl ? originalUri :
-                            this.servicePath + originalUri;
+                    String uriForSigning = presignedUrl ? uri :
+                            this.servicePath + uri;
                     expectedSignature = AwsSignature
                             .createAuthorizationSignatureV4(// v4 sign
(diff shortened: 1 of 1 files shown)


src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
```diff
@@ -630,8 +630,8 @@ public final void doHandle(HttpServletRequest baseRequest,
                         is = new ByteArrayInputStream(payload);
                     }
 
-                    String uriForSigning = presignedUrl ? originalUri :
-                            this.servicePath + originalUri;
+                    String uriForSigning = presignedUrl ? uri :
+                            this.servicePath + uri;
```
gaul, line 634: I am happy to merge this but is there any kind of test that you can add that failed before and succeeds now?

gaul (approved):

src/main/java/org/gaul/s3proxy/S3ProxyHandler.java
re1709: This is going to more complicated than thought to fix, to work with path style and virtual host style buckets. We've now switched to use host style buckets and no longer need the service path - everything works great! If i find some time i'll try to correct this change to not break host style.

Would it be feasible to block setting service path and virtual host at the same time in the S3Proxy builder? The two properties seem to conflict. Could then reliably make a change that checks one of them to not break functionality for the other.

PR #825, opened 2025-06-11
Fix EncryptedBlobStore support for s3 compatible backends without eTag parts number suffix

polarctos:
Some mostly S3 REST API compatible storage backends do not return the number of multipart upload parts as a suffix to the eTag as Amazon does and as the previous code expects. An example for this is [NetApp StorageGRID S3 REST API](https://docs.netapp.com/us-en/storagegrid/s3/s3-rest-api-supported-operations-and-limitations.html).

The old code had a fallback to just assume one encrypted part, but this is actually wrong in the multipart upload case for these backends, thus this problematic behaviour is replaced in this commit.

Example eTag structure from real AWS S3 for multipart uploads with suffix for 2 parts: `xyzabc-2`

In this case the number of parts is also not present in the object metadata, as metadata can only be set on S3 API when starting the multipart upload, and not when completing it, thus the number of parts is not yet known at this earlier point in time.

This change adds a third fallback option to just read the number of encrypted parts from the final part padding, which is the only place guaranteed to always be present for encrypted blobs.

Only for these backends this adds an additional GET request to the backend, but only for the actual 64 bytes of the padding. Before this change, the returned sizes were not correct for multipart uploads with more than one part on these backends without eTag parts number suffix.

### DIFF
--- a/src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
+++ b/src/main/java/org/gaul/s3proxy/EncryptedBlobStore.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.security.spec.KeySpec;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


polarctos:
> I can't look at this right now but does the undocumented S3 API to return the number of parts help here?
> 
> https://github.com/gaul/undocumented-s3-apis?tab=readme-ov-file#get-object-by-multipart-number

Unfortunately the `x-amz-mp-parts-count` response header is also not present from thios backend. There is only:
````
x-amz-request-id: 1749728871401780
x-amz-id-2: 12575733
````

As neither this additional header nor the eTag suffix are really specified by the [S3 API common response headers](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html) I guess there can also be more "mostly S3 compatible" storages that don't have these for multipart uploads. 
As the number of parts is also in the last part padding too for encrypted blobs, which is just a GET request for the last 64 bytes, I think thats acceptable as an additional GET request for such backends to make the encryption middleware compatible with these too. 

I rebased on master, to get the Python yield CI fix, could you please trigger to run the GitHub Actions again?

gaul:
Thank you for your contribution @polarctos!

PR #596, opened 2024-01-11
create docker images for armv7

snpz:
When by #595 pushing images to dockerhub is working again, this fixes #591 and creates an additional docker image for architecture linux/arm/v7.

PR #826, opened 2025-06-11
Run in Docker container on Eclipse Temurin OpenJDK 21 instead of 17

polarctos:
OpenJDK 21 is the current LTS version

This provides for example improved GC performance compared to OpenJDK 17.

This only increases the runtime JRE version for the provided Docker container, thus the Java code and build itself stays with compatibility on version 11 but at runtime the newer JVM benefits can still already be used.

https://endoflife.date/eclipse-temurin

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -175,7 +175,7 @@ jobs:
         uses: docker/build-push-action@v6
         with:
           context: .
-          platforms: linux/amd64,linux/arm64,linux/arm/v7
+          platforms: linux/amd64,linux/arm64
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ needs.meta.outputs.container_tags }}
           labels: ${{ needs.meta.outputs.container_labels }}


gaul:
CI fails due to some Python version change:
```
   ____________ ERROR collecting s3tests/functional/test_s3_website.py ____________
  'yield' keyword is allowed in fixtures, but not in tests (test_routing_generator)
```

polarctos:
> This PR should pass after rebasing to include [17e490e](https://github.com/gaul/s3proxy/commit/17e490e7759190709b3f83e2fb615b8588c5281e).

I now rebased on lastest master including the Python yield CI fix.

gaul:
Failed with:
```
Error: buildx failed with: ERROR: failed to solve: docker.io/library/eclipse-temurin:21-jre: failed to resolve source metadata for docker.io/library/eclipse-temurin:21-jre: no match for platform in manifest: not found
```

gaul:
> (A Remove `arm/v7` pre-built container

@snpz added armv7 support in #596.  Do you still need this or can you move to arm64?  I am skeptical that S3Proxy should support something that the OpenJDK packagers don't want to support.  I prefer dropping armv7.

PR #827, opened 2025-06-11
Add .dockerignore

polarctos:
Exclude everything from Docker build context that is not used by COPY steps in the Dockerfile.

This speeds up the container build on macOS, Windows or remote builders as less files and directories are copied over to the container builder. 

https://docs.docker.com/build/concepts/context/#dockerignore-files


gaul:
Thank you for your contribution @polarctos!

PR #828, opened 2025-06-12
Run on Ubuntu OpenJDK 21 JRE

polarctos:
OpenJDK 21 is the current LTS version. This provides for example improved GC performance compared to OpenJDK 17.

This only increases the runtime JRE version for the provided Docker container, thus the Java code and build itself stays with compatibility on version 11 but at runtime the newer JVM benefits can still already be used.

Alternative to https://github.com/gaul/s3proxy/pull/826 where `linux/arm/v7` would be dropped, this one retains all current architectures but switches the OpenJDK build from Temurin to Ubuntu builds.


gaul:
Obsoleted by #826.

PR #836, opened 2025-07-10
Add latency blobstore middleware

Phantom-101:
Allows running performance tests with s3proxy using simulated latency and read/write speed
- Added LatencyBlobStore, ThrottledInputStream
- Parse properties from s3proxy.conf
```
s3proxy.latency-blobstore.*.latency=1000 # Add 1000ms latency to all operations
s3proxy.latency-blobstore.put.latency=2000 # Override put blob to 2000ms latency
s3proxy.latency-blobstore.put.speed=10 # Limit put blob to 10 kb/s
```

### DIFF
--- a/src/main/java/org/gaul/s3proxy/LatencyBlobStore.java
+++ b/src/main/java/org/gaul/s3proxy/LatencyBlobStore.java
@@ -16,15 +16,8 @@
 
 package org.gaul.s3proxy;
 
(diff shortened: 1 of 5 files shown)


gaul:
Thank you for your contribution @Phantom-101!  Can you add a section to the README and a wiki page like other middlewares?

PR #851, opened 2025-08-28
Fix EncryptedBlobStore error handling

polarctos:
Fix for https://github.com/gaul/s3proxy/issues/848

The exceptions caught in `EncryptedBlobStore` were too broad, so that the jclouds `HttpResponseException` was not proxied to the client e.g. in a HTTP 304 conditional request response case.

Refactored the catch blocks to catch only the more specific exceptions related to the actual encryption.

The general `S3ProxyHandlerJetty` handler then handles the others:
https://github.com/gaul/s3proxy/blob/master/src/main/java/org/gaul/s3proxy/S3ProxyHandlerJetty.java#L88


polarctos:
I'll add a test in `EncryptedBlobStoreTest.java` too.

PR #904, opened 2025-10-18
Implement multipart upload for azureblob-sdk provider

klaudworks:
### Current Approach

The current approach forwards the input stream from the incoming request to the azure sdk. I used a `Flux<ByteBuffer>` to work around the limitation that our input stream does not support mark / reset. I looked into the azure-sdk-for-java after seeing your issue: https://github.com/Azure/azure-sdk-for-java/issues/42603. It seems like this is the only workaround.

### Current Limitations

1. Given, that we don't really have a place to store md5-based Etags for each part, we return them but don't actually use them to complete the multipart upload. Instead, we try to find all parts belonging to a multipart upload by using the upload id. For the finally assembled blob, we just return azure's Etag.
2. We can't make use of the azure-java-sdk retry logic on network issues. There is no way to solve this without persisting the input stream. The problem is pushed down to the s3-proxy client's aws sdk, which will retry uploading failed parts.
3. ~~We don't handle if-match, if-none-match headers for conditional writes. This is supported by S3 for CompleteMultipartUpload requests.~~ EDIT: I added a commit to support conditional writes for CompleteMultipartUpload requests.
4. AWS supports up to 5GiB part size. I updated the max part size to 4000 MiB as documented here: https://learn.microsoft.com/en-us/azure/storage/blobs/scalability-targets. The default in the AWS SDK is 8MB so I would rather not implement some complicated chunking of parts on our side to cover the edge case where people upload 4-5 GiB files. 

#### Using a BufferedInputStream 

I discarded this solution due to potentially high memory usage when uploading multiple parts in parallel.

#### Persisting the input stream on disk

Limitation 1: Potentially doubles the time until the part upload completes -> Can be compensated by uploading more files in parallel.

Limitation 2: Need to inform users that they provide sufficient /tmp storage on sufficiently fast SSDs. On an HDD the user would be heavily IO bound.

Benefit: We could actually calculate proper Etags for each part and store the md5 hash, e.g. encoded in the block name. This is not possible with the current solution because we need to already provide the block name when we pass along the input stream from the request to the azure sdk.

### s3-tests Update

https://github.com/gaul/s3-tests/pull/4

EDIT: I also running this on a test kubernetes cluster. It seems to work just fine. Some tooling (cnpg, kafka connector) can run their backups through it.

### Fixes

https://github.com/gaul/s3proxy/issues/709 https://github.com/gaul/s3proxy/issues/553 https://github.com/gaul/s3proxy/issues/552

### DIFF
--- a/src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
+++ b/src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
@@ -66,6 +66,9 @@ import com.azure.storage.common.policy.RetryPolicyType;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hashing;
+import com.google.common.hash.HashingInputStream;
 import com.google.common.io.BaseEncoding;
 import com.google.common.net.HttpHeaders;
 
@@ -117,6 +120,13 @@ import reactor.core.publisher.Flux;
 public final class AzureBlobStore extends BaseBlobStore {
     private static final String STUB_BLOB_PREFIX = ".s3proxy/stubs/";
     private static final String TARGET_BLOB_NAME_TAG = "s3proxy_target_blob_name";
+    private static final HashFunction MD5 = Hashing.md5();
+    // Disable retries since client should retry on errors.
+    private static final RequestRetryOptions NO_RETRY_OPTIONS = new RequestRetryOptions(
+            RetryPolicyType.FIXED, /*maxTries=*/ 1,
+            /*tryTimeoutInSeconds=*/ (Integer) null,
+            /*retryDelayInMs=*/ null, /*maxRetryDelayInMs=*/ null,
+            /*secondaryHost=*/ null);
 
     private final BlobServiceClient blobServiceClient;
     private final String endpoint;
@@ -133,11 +143,6 @@ public final class AzureBlobStore extends BaseBlobStore {
         this.endpoint = provider.getEndpoint();
         this.creds = creds;
         var cred = creds.get();
-        var retryOptions = new RequestRetryOptions(
-                RetryPolicyType.FIXED, 1,
-                (Integer) null,
-                null, null,
-                null);
         var blobServiceClientBuilder = new BlobServiceClientBuilder();
         if (!cred.identity.isEmpty() && !cred.credential.isEmpty()) {
             blobServiceClientBuilder.credential(
@@ -148,7 +153,7 @@ public final class AzureBlobStore extends BaseBlobStore {
         }
         blobServiceClient = blobServiceClientBuilder
                 .endpoint(endpoint)
-                .retryOptions(retryOptions)
+                .retryOptions(NO_RETRY_OPTIONS)
                 .buildClient();
     }
 
@@ -842,10 +847,11 @@ public final class AzureBlobStore extends BaseBlobStore {
                 mpu.containerName(), mpu.blobName());
 
         byte[] md5Hash;
-        try {
-            var is = payload.openStream();
+        try (var is = payload.openStream();
+             var his = new HashingInputStream(MD5, is)) {
             var providedMd5 = payload.getContentMetadata().getContentMD5AsHashCode();
-            var md = MessageDigest.getInstance("MD5");
+
+            final int maxChunkSize = 4 * 1024 * 1024;
 
             Flux<ByteBuffer> body = Flux.generate(
                 () -> 0L,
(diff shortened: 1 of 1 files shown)


src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
```diff
@@ -571,65 +584,436 @@ public void setBlobAccess(String container, String key, BlobAccess access) {
...
+                () -> 0L,
+                (position, sink) -> {
+                    try {
+                        if (position >= contentLength) {
+                            sink.complete();
+                            return position;
+                        }
+                        int chunkSize = (int) Math.min(4 * 1024 * 1024L,
+                                contentLength - position);
+                        ByteBuffer buffer = ByteBuffer.allocate(chunkSize);
```
gaul: Can you allocate this outside the loop and reuse it within the loop?

src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
```diff
@@ -571,65 +584,436 @@ public void setBlobAccess(String container, String key, BlobAccess access) {
...
+                position -> {
+                    try {
+                        is.close();
+                    } catch (IOException e) {
+                        // Ignore close errors
+                    }
+                }
+            );
+
+            asyncClient.stageBlock(blockId, body, contentLength).block();
```
gaul: Thanks for finding that the async client allows streaming writes!  I did not anticipate that Microsoft would support this in async but not in the sync client.

src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
```diff
@@ -118,13 +131,13 @@ public final class AzureBlobStore extends BaseBlobStore {
...
-        // Disable retries since client should retry on errors.
         var retryOptions = new RequestRetryOptions(
-                RetryPolicyType.FIXED, /*maxTries=*/ 1,
-                /*tryTimeoutInSeconds=*/ (Integer) null,
-                /*retryDelayInMs=*/ null, /*maxRetryDelayInMs=*/ null,
-                /*secondaryHost=*/ null);
+                RetryPolicyType.FIXED, 1,
+                (Integer) null,
+                null, null,
+                null);
```
gaul: Please revert these noisy changes.

src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
```diff
@@ -571,65 +584,436 @@ public void setBlobAccess(String container, String key, BlobAccess access) {
...
+    /**
+     * Creates a BlockBlobAsyncClient with retries disabled for streaming uploads.
+     * This allows us to stream directly from non-markable InputStreams without
+     * needing temp files or buffering. The S3 client can retry the entire part
+     * upload if needed.
+     */
+    private BlockBlobAsyncClient createNonRetryingBlockBlobAsyncClient(
+            String container, String blobName) {
+        var cred = creds.get();
+        var retryOptions = new RequestRetryOptions(
```
gaul: Could you hoist this into a static final and use the options for both the sync and async clients?

src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
```diff
@@ -571,65 +584,436 @@ public void setBlobAccess(String container, String key, BlobAccess access) {
...
+                (position, sink) -> {
+                    try {
+                        if (position >= contentLength) {
+                            sink.complete();
+                            return position;
+                        }
+                        int chunkSize = (int) Math.min(4 * 1024 * 1024L,
+                                contentLength - position);
+                        ByteBuffer buffer = ByteBuffer.allocate(chunkSize);
+                        int read = is.read(buffer.array(), 0, chunkSize);
```
gaul: I believe that a call to `InputStream.read` may return fewer bytes than requested yet more may come from a socket in a subsequent call.  So you really want a loop:

```java
for (int i = 0; i < chunkSize; ) {
    int read = is.read(buffer.array(), i, buffer.remaining());
    if (read == -1) { ... }
    i += read;
    ...
}
```

gaul (changes requested):
Please address comments in `uploadMultipartPart` but overall this looks good!

src/main/java/org/gaul/s3proxy/azureblob/AzureBlobStore.java
klaudworks: this may be a bad idea because there is no guarantee that the AzureSDK strictly processes one ByteBuffer after another. Instead, we may overwrite a ByteBuffer that the AzureSDK still relies on.

DzeCin:
> @DzeCin what makes you think so? There should be no state persisted in the s3 proxy.

You are right I misread the code, sorry for the confusion.

klaudworks:
@gaul no hurry, I am currently running a custom build in production anyways. Thank you for reviewing the changes!

PR #994, opened 2026-03-04
Sanitize TARGET_BLOB_NAME_TAG values on Azure

miselin:
Blob names have a wider set of permitted characters than tag values, so it's possible for the target blob name tag to cause an error even though the blob itself will create successfully.

This will now base64-encode the blob name to ensure it's always writable.

This would be a smaller patch, but I wanted to ensure there are two backwards-compatible safeties:
1. Values without the base64 marker will continue to decode correctly
2. Values with the base64 marker but without an internal marker will decode as-is, which handles the case where an s3proxy client happens to use the same base64 marker string for some reason

Tested via `mvn test`, and by running a custom build in production to confirm the fix works.


gaul:
Thanks for raising this issue!  The original code should have stored the target blob name in user metadata instead of tags.

PR #1026, opened 2026-04-15
Close var in testBlobNotEncrypted

tejasae-afk:
While tracing testBlobNotEncrypted in src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java I noticed what might be an issue: The resource opened there can leak if testBlobNotEncrypted exits on an error path. this patch moves the allocation into try-with-resources so cleanup happens on every exit path. Let me know if the approach doesn’t fit.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.

### DIFF
--- a/src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java
+++ b/src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java
@@ -169,24 +169,22 @@ public final class EncryptedBlobStoreTest {
             blobStore.putBlob(containerName, blob);
             blob = encryptedBlobStore.getBlob(containerName, blobName);
 
-            InputStream blobIs = blob.getPayload().openStream();
-            var r = new InputStreamReader(blobIs);
-            var reader = new BufferedReader(r);
-            String plaintext = reader.lines().collect(Collectors.joining());
-            logger.debug("plaintext {}", plaintext);
-
-            assertThat(content).isEqualTo(plaintext);
+            try (InputStream blobIs = blob.getPayload().openStream()) {
+                var reader = new BufferedReader(new InputStreamReader(blobIs));
+                String plaintext = reader.lines().collect(Collectors.joining());
+                logger.debug("plaintext {}", plaintext);
+                assertThat(content).isEqualTo(plaintext);
+            }
 
             var options = new GetOptions();
             blob = encryptedBlobStore.getBlob(containerName, blobName, options);
 
-            blobIs = blob.getPayload().openStream();
-            r = new InputStreamReader(blobIs);
-            reader = new BufferedReader(r);
-            plaintext = reader.lines().collect(Collectors.joining());
-            logger.debug("plaintext {} with empty options ", plaintext);
-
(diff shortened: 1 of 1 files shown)


src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java
```diff
@@ -170,23 +170,25 @@ public void testBlobNotEncrypted() throws Exception {
             blob = encryptedBlobStore.getBlob(containerName, blobName);
 
             InputStream blobIs = blob.getPayload().openStream();
-            var r = new InputStreamReader(blobIs);
-            var reader = new BufferedReader(r);
-            String plaintext = reader.lines().collect(Collectors.joining());
-            logger.debug("plaintext {}", plaintext);
+            try (var r = new InputStreamReader(blobIs)) {
```
gaul: This pattern repeats several times -- want to fix this all at once?

gaul: Also you should be using try-with-resources on `blobIs` not `r`

src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java
```diff
@@ -170,23 +170,25 @@ public void testBlobNotEncrypted() throws Exception {
...
+                    blobIs = blob.getPayload().openStream();
+                    r = new InputStreamReader(blobIs);
+                    reader = new BufferedReader(r);
+                    plaintext = reader.lines().collect(Collectors.joining());
+                    logger.debug("plaintext {} with empty options ", plaintext);
 
-            assertThat(content).isEqualTo(plaintext);
+                    assertThat(content).isEqualTo(plaintext);
+            }
+            }
```
gaul, line 187: Extra brace?

tejasae-afk:
Good catches, thanks. Updated:

- Wrapped `blobIs` (not `r`) in try-with-resources throughout, so the `InputStream` itself is the resource being managed rather than the thin `InputStreamReader` wrapper over it.
- Applied the same fix to all occurrences across the file: `testBlobNotEncrypted`, `testBlobNotEncryptedRanges`, `testEncryptContent`, `testEncryptContentWithOptions`, `testEncryptMultipartContent`, `testReadPartial`, `testReadTail`, `testReadPartialWithRandomEnd`, `testMultipartReadPartial`, `testMultipartReadTail`, and `testMultipartReadPartialWithRandomEnd`.
- Removed the extra brace.

src/test/java/org/gaul/s3proxy/EncryptedBlobStoreTest.java
tejasae-afk: makes sense!

gaul:
> Good catches, thanks. Updated:

Did you forget to push a new commit?

tejasae-afk:
Sorry about that — commit is up now.

gaul:
Thank you for your contribution @tejasae-afk!

PR #1042, opened 2026-05-06
Bridge S3 backup clients to SFTP storage

captainpacket:
## Summary

Adds an SFTP-backed BlobStore provider so S3-compatible clients can write through S3Proxy to SFTP storage.

## Details
- Reuses the existing NIO.2 BlobStore implementation over Apache MINA SSHD SFTP.
- Keeps S3Proxy client credentials separate from SFTP backend credentials.
- Requires a pinned SFTP host-key fingerprint via `jclouds.sftp.host-key`.
- Maps objects to `<jclouds.sftp.basedir>/<bucket>/<object-key>`.
- Closes the SFTP FileSystem and SSH client through the jclouds context lifecycle.

## Validation
- `mvn verify`
- `mvn test -Dtest=SftpBlobStoreTest,S3BackupSftpBridgeTest`

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Maven Central hosts S3Proxy artifacts and the wiki has
 * rackspace-cloudfiles-uk and rackspace-cloudfiles-us
 * s3 (non-Amazon, alias for aws-s3-sdk)
 * s3-jclouds (non-Amazon S3 via jclouds, deprecated)
+* sftp (SFTP storage via Apache MINA SSHD)
 * transient (in-memory storage, alias for transient-nio2)
 * transient-nio2 (in-memory storage, recommended)
 * transient-jclouds (in-memory storage via jclouds, deprecated)
--- a/docs/SFTP.md
+++ b/docs/SFTP.md
@@ -1,0 +1,98 @@
+# SFTP
+
+Status: experimental
+
+The SFTP backend exposes an SFTP server as an S3Proxy storage backend through
+the jclouds `BlobStore` interface. It reuses the existing NIO.2 blobstore
+implementation over Apache MINA SSHD's SFTP filesystem provider.
+
+## Security
+
+The SFTP backend requires a pinned SFTP server host-key fingerprint. Set
+`jclouds.sftp.host-key` to the expected fingerprint, for example
+`SHA256:...`. S3Proxy rejects the SFTP connection if the server presents a
+different host key.
+
+Keep S3Proxy client credentials separate from SFTP credentials. S3 clients
+authenticate to S3Proxy with `s3proxy.identity` and `s3proxy.credential`; S3Proxy
+authenticates to the SFTP server with `jclouds.identity` and
+`jclouds.credential`.
+
+## Configuration
+
+Example:
+
+```
+s3proxy.authorization=aws-v2-or-v4
+s3proxy.endpoint=http://127.0.0.1:8080
+s3proxy.identity=local-identity
+s3proxy.credential=local-credential
+
+jclouds.provider=sftp
+jclouds.endpoint=sftp://127.0.0.1:2222/
+jclouds.identity=sftp-user
+jclouds.credential=sftp-password
+jclouds.sftp.basedir=/s3proxy
+jclouds.sftp.host-key=SHA256:...
+```
+
+If `jclouds.endpoint` omits a port, the backend uses SFTP port 22. The default
+`jclouds.sftp.basedir` is `/s3proxy`.
+
+You can usually get an OpenSSH-format SHA256 host-key fingerprint with:
+
+```
+ssh-keyscan -p 2222 127.0.0.1 | ssh-keygen -lf -
+```
+
+## Storage Mapping
+
+S3 buckets are first-level directories below `jclouds.sftp.basedir`. Object keys
+map to files below the bucket directory. The effective path mapping is:
+
+```
+<jclouds.sftp.basedir>/<bucket>/<object-key>
+```
+
+Example:
+
+```
+jclouds.sftp.basedir=/data/backups
+bucket=example-bucket
+object-key=backups/app/1000/db_dump.zip
+```
+
+Stored SFTP path:
+
+```
+/data/backups/example-bucket/backups/app/1000/db_dump.zip
+```
+
+The bucket name is therefore part of the SFTP path. For S3-compatible backup
+clients, configure the client bucket name to the directory name you want under
+`jclouds.sftp.basedir`.
+
+## Metadata
+
+The NIO.2 backend stores content metadata and user metadata in user extended
+attributes when the filesystem supports them. SFTP servers generally do not
+provide portable user extended attributes, so the SFTP backend treats them as
+optional.
+
+Object bytes are stored as regular SFTP files. Object size and last-modified
+values come from SFTP file attributes. S3 user metadata is not portable across
+SFTP servers.
+
+## Compatibility Profile
+
+The embedded SFTP test covers the common S3-compatible backup operation set:
+
+* create bucket;
+* upload payload and metadata objects;
+* `HEAD` payload objects;
+* `GET` payload and metadata objects;
+* list objects with a prefix;
+* delete objects and bucket.
+
+Broader S3 API compatibility should be validated against the target SFTP server
+before production use.
--- a/pom.xml
+++ b/pom.xml
@@ -424,6 +424,7 @@
     <opentelemetry-semconv.version>1.41.0</opentelemetry-semconv.version>
     <slf4j.version>2.0.17</slf4j.version>
     <shade.prefix>${project.groupId}.shaded</shade.prefix>
+    <sshd.version>2.17.1</sshd.version>
     <surefire.version>3.5.5</surefire.version>
   </properties>
 
@@ -563,6 +564,11 @@
       <groupId>org.apache.jclouds.driver</groupId>
       <artifactId>jclouds-slf4j</artifactId>
       <version>${jclouds.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.sshd</groupId>
+      <artifactId>sshd-sftp</artifactId>
+      <version>${sshd.version}</version>
     </dependency>
     <dependency>
       <groupId>org.assertj</groupId>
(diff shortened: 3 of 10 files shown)


gaul:
Thanks for the PR — overall shape (extending the NIO.2 abstraction over MINA SSHD's SFTP FileSystemProvider) reuses the right infrastructure. A few things I'd want addressed before this lands.

### Blocking

1. **`LocalS3BackupClientSmokeTest.java` is out of scope.** It logs into some app at `https://localhost:8443` with CSRF tokens and POSTs to `/api/backup-settings` and `/api/backup-settings/storage`. Those endpoints are bespoke to a specific external application — not S3, not a general profile. Even though it's env-var gated, 379 lines of code in s3proxy testing a third-party UI doesn't belong here. Please remove.

2. **SFTP `FileSystem` is never closed.** In `SftpBlobStore.createRoot`:
   ```java
   var fs = new SftpFileSystemProvider().newFileSystem(uri, Map.of());
   var root = fs.getPath(baseDir).normalize();
   ```
   The `FileSystem` holds the SSH session. Nothing wires it into `BlobStoreContext.close()`, so every context rebuild leaks an SSH connection. `AbstractNio2BlobStore` doesn't manage `fs` lifecycle today (see the `// TODO: close fs?` in `TransientNio2BlobStore`), but that becomes load-bearing here because SFTP connections aren't free.

3. **Port doesn't default to 22.** `URI.create("sftp://host/").getPort()` returns `-1`, which is passed straight to `SftpFileSystemProvider.createFileSystemURI(host, -1, ...)`. The default endpoint includes `:22` so the happy path works, but `jclouds.endpoint=sftp://example.com/` will fail. `port < 0 ? 22 : port`.

4. **No host key verification.** MINA SSHD's default `ServerKeyVerifier` accepts whatever the server offers. For a proxy translating S3 traffic to an upstream SFTP host, that's a serious MITM gap on the SFTP leg. The doc mentions this as future work, but at minimum, document the threat prominently and ideally add basic host-key pinning before un-drafting.

5. **Hardcoded `sshd-sftp` version `2.17.1`.** Rest of `pom.xml` uses `${name.version}` properties — please follow suit with `${sshd.version}`. Also verify test-only `SshServer`/`SftpSubsystemFactory` classes resolve without needing an explicit `sshd-core` test dep.

### Design / correctness

6. `ConsistencyModel.STRICT` is wrong for a remote SFTP target. Existing nio2 backends are local; SFTP is not.

7. `BASEDIR` default of `/` is a foot-gun (buckets at SFTP root). Either require explicit setting or default to something like `/s3proxy`.

8. Credentials with URI-special chars (`@`, `:`, `/`, `#`, `%`) passed to `createFileSystemURI` — please add a unit test with e.g. a password containing `@`. Either it works (great) or it's a silent auth failure.

9. `Files.createDirectories(root)` on `/` may be rejected by some SFTP servers at the chroot. Tolerate `FileAlreadyExistsException` / skip when path exists.

### Tests

10. **`S3BackupSftpBridgeTest` coverage gap vs spec.** Docs say HEAD/GET the payload objects (plural); the test only HEADs `snapshotPayload` and only GETs `entry.txt` text files. Add HEAD on `appPayload` and content GETs on both payloads to verify body roundtrip.

11. **`Files.exists(sftpRoot.resolve(...))` assertions** test the fixture filesystem, not the SFTP layer. The S3 GET roundtrip already proves storage; these exist-checks just couple the test to `VirtualFileSystemFactory`'s mapping. Drop or replace with content checks.

12. **Temp resource leaks.** `Files.createTempDirectory("s3proxy-sftp-root")` and the hostkey temp file are never cleaned up in `tearDown`. JUnit `@TempDir` or explicit recursive delete.

13. **`SftpBlobStoreTest` only checks `ContextBuilder.newBuilder("sftp") != null`** — that just verifies `@AutoService` ran. Either expand or drop.

14. **JUnit 4 vs 5 mix.** Existing `AwsSdkTest` etc. use JUnit 5; new tests use JUnit 4. Pick one.

### Docs / packaging

15. README "Supported storage backends" list isn't updated — add `sftp`.

16. `docs/backends/sftp.md` references `src/test/resources/s3proxy-sftp.conf` and an `AwsSdkTest`/`run-s3-tests.sh` lane that don't exist. Either add the conf and wire the test lane in, or drop the references.

17. Existing docs are flat (`docs/Encryption.md`, `docs/Logging.md`); PR adds nested `docs/backends/` and `docs/compatibility/`. Either move the old ones for consistency or keep new files flat.

18. Status strings disagree: backends doc says "S3 backup profile validated", compatibility doc says "embedded SFTP profile validated". Pick one.

### Nits

- `throw new RuntimeException("Failed to initialize SFTP backend", ioe)` — consider a more specific exception so init failures are distinguishable.
- `getXattrView` swallowing `UnsupportedOperationException` per call is correct, but logging "xattrs unavailable" once at provider init would be cleaner than per-write debug lines.

### Suggested order

1. Drop `LocalS3BackupClientSmokeTest`.
2. Fix FileSystem lifecycle and port default.
3. Add `s3proxy-sftp.conf` and parameterize `AwsSdkTest` over it to get real S3 compat coverage (not just the backup profile).
4. README, version property, test cleanup, JUnit consistency.
5. Document or implement host-key verification before un-drafting.

captainpacket:
Addressed the review items in the amended single commit:

- removed the application-specific local smoke test;
- wired the SFTP `FileSystem` into jclouds `Closer` so `BlobStoreContext.close()` closes the SSH-backed filesystem;
- defaulted omitted SFTP endpoint ports to 22;
- documented the current host-key verification limitation prominently in flat `docs/SFTP.md`;
- moved `sshd-sftp` to `${sshd.version}`;
- changed SFTP consistency to `EVENTUAL` and default `jclouds.sftp.basedir` to `/s3proxy`;
- added credential-special-character and default-port coverage;
- converted new tests to JUnit 5 with `@TempDir` cleanup;
- expanded the backup profile test to HEAD and GET both payload objects, and removed fixture filesystem assertions;
- updated README supported backends and removed nested/missing-reference docs.

Validation: `mvn verify` passes locally.

captainpacket:
Added host-key pinning on top of the previous cleanup:

- new required `jclouds.sftp.host-key` property;
- SFTP connections now use a MINA `ServerKeyVerifier` backed by the pinned fingerprint;
- the SSH client is closed through the same jclouds lifecycle path as the SFTP `FileSystem`;
- tests cover both a matching embedded server key and a mismatched-host-key rejection;
- docs now show how the host-key fingerprint maps into SFTP configuration.

Validation: `mvn verify` passes locally.

captainpacket:
Thanks for the detailed review here. I took the host-key concern as worth fixing now rather than leaving as documentation-only, so the branch now requires a pinned SFTP host-key fingerprint and rejects mismatches in the embedded SFTP test.

I also tried to keep the rest of the cleanup narrow: no app-specific smoke path, flat docs, lifecycle cleanup, and the existing upstream `mvn verify` gate passing locally. Happy to adjust the shape if you would prefer broader S3 compatibility coverage before this comes out of draft.

gaul (commented):
## Overall

Solid architecture — reusing `AbstractNio2BlobStore` over SshD's `SftpFileSystemProvider` is the right approach, and pinned host-key verification (no opportunistic TOFU) is the right security default. Code style matches the existing filesystem/transient-nio2 providers. "Experimental" status in the docs is appropriate.

## Findings

**Resource cleanup gap in `createRoot` (SftpBlobStore.java:282-303)**
The catch blocks only close the `SshClient`. If `Files.createDirectories(root)` throws after `newFileSystem(...)` succeeds, the `FileSystem` is orphaned. Closing the client typically closes the SSH session, but the `FileSystem` instance is registered with `SftpFileSystemProvider` and remains in its tracking map. Suggest closing `fs` first, then the client, on the failure paths.

**`getXattrView` catches `UnsupportedOperationException` but call sites further down don't (AbstractNio2BlobStore.java:1186-1208)**
After `getXattrView` returns a non-null view on an SFTP-backed path, `view.list()` / `view.write()` / `view.read()` could still throw `UnsupportedOperationException`. Existing `safeGetXattrs` only catches `IOException`. If SshD always returns null/throws from `getFileAttributeView` itself (which the comment suggests), this is fine — but worth a one-line comment or an integration test that triggers it.

**`closeClient(Channel client, ...)` (SftpBlobStore.java:349)**
`Channel` is an odd parameter type here — `Closeable` is the conventional choice and doesn't require importing `java.nio.channels.Channel`. SshClient satisfies both since SshD's `Closeable` extends `Channel`. Cosmetic.

**`createFileSystemUri` with null identity (SftpBlobStore.java:326-336)**
If `creds.identity` is null (no credentials configured), the userInfo becomes the string `\"null\"` or `\"null:...\"`. Worth either rejecting this earlier with a clear error or treating null identity as anonymous.

**Test: `testRejectsMismatchedHostKey` (SftpBlobStoreTest.java:892-899)**
Verified this does exercise the verifier: `BlobStoreContextImpl`'s constructor takes `BlobStore` directly, so `ContextBuilder.build()` eagerly instantiates `SftpBlobStore`, which connects and triggers host-key verification. Test is correct, but consider tightening the assertion from `RuntimeException` to something more specific (or asserting on the cause/message) so a regression that fails for the *wrong* reason doesn't go unnoticed.

**Test scope**
- No test for invalid SFTP credentials (auth failure path).
- No test for the xattr-unsupported fallback the PR adds to `AbstractNio2BlobStore`. The SshD virtual FS test server likely doesn't exercise the `UnsupportedOperationException` branch — confirmed by inspection that the helper is reached via `getXattrView` but not validated end-to-end.
- `S3BackupSftpBridgeTest` doesn't exercise \`CopyObject\` or AWS-style multipart upload (1MB/2MB payloads stay below the SDK's default multipart threshold even with \`multipartEnabled(true)\`). For \"backup client\" coverage these are common operations worth adding.

**Minor**
- \`SftpBlobStoreApiMetadata.defaultEndpoint(\"sftp://127.0.0.1:22/\")\` — fine, but the README mentions port 22 default while the example doc uses 2222. Consistent enough.
- The two-step constructor + \`Root\` record is clean — no objection, though a single-constructor variant with helper methods would also work.

Nothing here is a blocker for the draft. The cleanup-on-failure issue is the most material; the rest are polish/coverage.

gaul:
Please resolve the merge conflict and address the review comments but I'll look at this once more when you mark it ready for review.

captainpacket:
Rebased on current `upstream/master` and addressed the latest review pass:

- resolved the README / `AbstractNio2BlobStore` merge conflicts;
- closed the SFTP `FileSystem` before closing the SSH client on initialization failure;
- switched the cleanup helper to `Closeable`;
- rejects missing SFTP identity with a clearer error before URI construction;
- handles `UnsupportedOperationException` from xattr list/write paths, not only from view lookup;
- tightened the mismatched-host-key test to assert the verifier failure path.

Validation: `mvn verify` passes locally on the rebased branch.

gaul:
Thank you for your contribution @captainpacket!

PR #1046, opened 2026-05-13
Remove jclouds dependency by forking the blobstore API

gaul:
Copy the jclouds blobstore types we use (BlobStore, Blob, BlobMetadata,
MultipartUpload, options, payloads, exceptions) into a new
org.gaul.s3proxy.blobstore package, strip Guice dependency injection,
and replace ContextBuilder with a direct BlobStores factory. Drop the
legacy jclouds-only providers (filesystem, transient, aws-s3, azureblob,
google-cloud-storage); only the SDK-backed providers remain
(filesystem-nio2, transient-nio2, aws-s3-sdk, azureblob-sdk,
google-cloud-storage-sdk).

The forked types are modernized rather than preserved verbatim:
BlobMetadata, ContainerMetadata, MultipartUpload, MultipartPart,
ContentMetadata, HttpResponse, the *Options classes, and Credentials
are records; StorageMetadata is a sealed interface; Blob and Payload
are immutable with Builder factories (Blob.Builder is nested and takes
the name at construction). Several redundant abstractions are dropped
along the way: BlobStore.blobBuilder, the Multimap-based allHeaders
(collapsed to a single Content-Range field), updateMetadata Consumer
overloads (replaced by direct setters), and Payload.release() (no
callers).


gaul:
The next release 4.0.0 will include this PR.

PR #1074, opened 2026-06-15
Fix AwsS3SdkBlobStore.containerExists to translate non-404 S3Exception

### LINKED ISSUE #1073: AwsS3SdkBlobStore.containerExists re-throws raw S3Exception for non-404 errors, causing 500 response
Version: 3.1.0

## Summary
When using the aws-s3-sdk backend against an S3-compatible endpoint (e.g. OpenStack Swift with s3-middleware), a HEAD /<container>/ request that returns 403 Forbidden for a non-existent container causes s3proxy to return 500 Internal Server Error instead of the expected 404 Not Found.

## Root cause
In AwsS3SdkBlobStore.containerExists, when the AWS SDK throws an S3Exception with a status code other than 404, the exception is re-thrown raw:

```java
} catch (S3Exception e) {
    if (e.statusCode() == 404) {
        return false;
    }
    throw e;  // raw S3Exception — not translated to a jclouds exception
}
```
S3ProxyHandler only handles jclouds exceptions. A raw S3Exception escapes all handlers and Jetty returns 500.

## Fix
Add `throw translate(e, container, null)` before `throw e` so the exception is converted to a jclouds ContainerNotFoundException or similar, which S3ProxyHandler handles correctly.

## Context
OpenStack Swift returns 403 Forbidden (not 404) when performing HEAD on a non-existent container, so this is reliably triggered during e.g. `mc alias set` probe requests against Swift-backed s3proxy instances. The jclouds s3 backend does not have this issue as it translates exceptions through its own error handling chain.

mmieszczak:
When containerExists receives an S3Exception with a status code other than 404, it re-throws it raw. S3ProxyHandler only handles jclouds exceptions, so the raw S3Exception propagates to Jetty and results in a 500 Internal Server Error.
Change:

```diff
+ throw translate(e, container, null);
  throw e;
```

### Why this matters
Some S3-compatible backends return 403 Forbidden instead of 404 Not Found for HEAD requests on non-existent containers. OpenStack Swift is one example — it returns 403 for a non-existent container when the requesting credentials don't have explicit list permissions. This causes every containerExists probe (e.g. during mc alias set) to produce a 500 instead of the correct 404/false response.
The jclouds s3 backend does not have this issue because all exceptions go through its own translation chain before reaching S3ProxyHandler

Fixes https://github.com/gaul/s3proxy/issues/1073

### DIFF
--- a/src/main/java/org/gaul/s3proxy/awssdk/AwsS3SdkBlobStore.java
+++ b/src/main/java/org/gaul/s3proxy/awssdk/AwsS3SdkBlobStore.java
@@ -288,7 +288,7 @@ public final class AwsS3SdkBlobStore extends BaseBlobStore {
             if (e.statusCode() == 404) {
                 return false;
             }
-            throw e;
+            throw translate(e, container, null);
         }
     }
 


gaul:
> Some S3-compatible backends return 403 Forbidden instead of 404 Not Found for HEAD requests on non-existent containers. OpenStack Swift is one example — it returns 403 for a non-existent container when the requesting credentials don't have explicit list permissions. This causes every containerExists probe (e.g. during mc alias set) to produce a 500 instead of the correct 404/false response.

This looks like AI gibberish -- it doesn't even compile.

mmieszczak:
Sorry, I have initially based on tag 3.1.0 instead of master and didn't know, that error wrapping pattern has changed... This should build now. This is a single line change to pass backend http codes to client instead of responding with 500.

gaul:
Thank you for your contribution @mmieszczak!

PR #1094, opened 2026-06-25
Report Size/LastModified/ETag in the ListObjects call for `directory-marker` objects stored nio2-based backends. Fixes #1093

### LINKED ISSUE #1093: Nio2 backends omit <Size>, <LastModified>, and <ETag> from ListObjects/ListObjectsV2 for directory placehoolder keys (ending in /), breaking integraiton with hadoop-aws.
Summary
  
  S3Proxy's nio2 backends (filesystem-nio2, transient-nio2) omit `<Size>`, `<LastModified>`, and `<ETag>` from ListObjects/ListObjectsV2 `<Contents>` entries for "directory placeholder" keys (keys ending in /), even though the object exists and HEAD returns its size.
  This breaks integration with hadoop-aws.

Problem

  Per the S3 API, every <Contents> entry must carry `<Size>` (plus `<LastModified>` and `<ETag>`). Omitting them makes spec-abiding clients read a null/absent size. Concretely, Hadoop's S3A connector then:
  - NPEs in getObjectSize() (S3Object.size() is null), and
  - fails getFileStatus(`<dir>`), so it cannot recognize the directory.
  
  This breaks HBase bulk load (StoreFileWriter: "Expecting <path> to be a directory") when backing up/restoring to S3Proxy over s3a://.

   Steps to reproduce

  1. Start S3Proxy with JCLOUDS_PROVIDER=filesystem-nio2.
  2. PUT two empty (0-byte) objects differing only by a trailing slash: dir-marker/ and regular-object.
  3. HEAD dir-marker/ → returns Content-Length: 0 (server knows the size).
  4. ListObjectsV2 / ListObjects.

  Expected

  The dir-marker/ `<Contents>` entry includes `<Size>0</Size>`, `<LastModified>`, and `<ETag>` — identical to the regular-object entry.

  Actual
  
  The dir-marker/ entry contains only `<Key>` and `<StorageClass>`; `<Size>`, `<LastModified>`, and `<ETag>` are missing. regular-object (same 0 bytes, no trailing slash) has all three.

```
  <Contents>
    <Key>dir-marker/</Key>                                                                                                                                                                                                                                                                                              
    <StorageClass>STANDARD</StorageClass>
  </Contents>
```

craffit:
This PR fixes and issue where `directory-marker` objects, which are empty objects with a trailing "/" do not get their object information reported for the ListObjects() call. This is a problem in hadoop-aws, which release on those kinds of objects.

More details:

AbstractNio2BlobStore.listHelper() added directory-marker objects (keys ending in "/", stored as a directory carrying the XATTR_CONTENT_MD5 xattr) to listings with hardcoded null eTag, lastModified and size. For a non-delimited ListObjects(V1/V2) the handler emits such an entry as `<Contents>` but then omits `<Size>`, `<LastModified>` and `<ETag>` because each is null, so spec-abiding clients (e.g. Hadoop's S3A connector) read a null size and NPE.

Populate the metadata from the directory's attributes when the marker object exists (eTag from XATTR_CONTENT_MD5, size 0, lastModified from the file), mirroring the getBlob/HEAD path. Implicit prefixes that have no marker object keep null metadata since they surface only as `<CommonPrefixes>`. Factor the duplicated ETag-from-xattr logic into a readETagXattr() helper.

This fixes https://github.com/gaul/s3proxy/issues/1093

### DIFF
--- a/src/main/java/org/gaul/s3proxy/nio2blob/AbstractNio2BlobStore.java
+++ b/src/main/java/org/gaul/s3proxy/nio2blob/AbstractNio2BlobStore.java
@@ -239,17 +239,40 @@ public abstract class AbstractNio2BlobStore extends BaseBlobStore {
                                 filterMultipart);
                     }
 
+                    var dirXattrs = safeGetXattrs(path);
+                    var markerExists = dirXattrs.attributes()
+                            .contains(XATTR_CONTENT_MD5);
+
                     // Add a prefix if the directory blob exists or if the delimiter causes us not to recuse.
-                    if ("/".equals(delimiter) || safeGetXattrs(path).attributes().contains(XATTR_CONTENT_MD5)) {
+                    if ("/".equals(delimiter) || markerExists) {
                         var name = relativeName(containerPath, path);
                         logger.debug("adding prefix: {}", name);
+
+                        // A directory-marker object (a key ending in "/") that
+                        // was explicitly stored carries the XATTR_CONTENT_MD5
+                        // xattr. Report its metadata so a non-delimited
+                        // ListObjects, which emits this entry as <Contents>,
+                        // includes Size/LastModified/ETag like any other
+                        // 0-byte object. Implicit prefixes (no marker object)
+                        // keep null metadata since they surface only as
+                        // <CommonPrefixes>.
+                        String eTag = null;
+                        Date lastModified = null;
+                        Long size = null;
+                        if (markerExists) {
+                            eTag = readETagXattr(dirXattrs);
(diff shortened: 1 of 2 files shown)


craffit:
Late addition, this also fixes an issue where an empty directory is not reported as a blob. This is also required by hadoop-aws (https://github.com/gaul/s3proxy/pull/1094/changes/7925896cf3f6d48cee3c9de5ae80cd5a24ddbdab)

gaul:
Thank you for your contribution @craffit!

PR #1130, opened 2026-08-06
Acknowledge SSE-S3 (accept + echo), partially resolving #402

### LINKED ISSUE #402: Object server-side encryption
Lack of support prevents some kinds of testing, e.g., s3fs integration tests with `-o use_sse`.  Full support for this requires jclouds work like apache/jclouds#130.  Is there some shortcut here instead of implementing a few providers and the portable abstraction?  Perhaps just passing through the S3 headers?  I notice that https://github.com/jiffysrc/s3proxy/commit/fc8547fb52c4013712760ee93b648d469fea5035 has some partial support.  @jiffysrc

gavin-thompson-postman:
Partially resolves #402 — this is **Phase 1 (SSE-S3 accept + echo)** from your comment on that issue. Opening as a draft to check the approach before going further.

## What this does
Adds an `s3proxy.server-side-encryption` property governing the `x-amz-server-side-encryption` request header:

- **Default reflects what the backend can truthfully claim** — `sse-s3` on stores that always encrypt at rest (S3 / Azure / GCS, via `Quirks.SERVER_SIDE_ENCRYPTION_AT_REST`) or when `s3proxy.encrypted-blobstore` is enabled; otherwise `reject`. This follows the honesty point in the issue ("gate the echo on a real backend or on `EncryptedBlobStore` being enabled rather than assert encryption that is not happening").
- **Overridable** via the property: `reject` | `ignore` | `sse-s3`.
- In `sse-s3` mode, `AES256` is accepted and **echoed on PUT / HEAD / GET / copy / complete**.
- `aws:kms` and SSE-C (`x-amz-server-side-encryption-customer-*`) are **always rejected** with `501`, so a caller-supplied key is never silently dropped.

The header is added to `SUPPORTED_X_AMZ_HEADERS` (building on the allowlist idea in jiffysrc@fc8547f) and handled explicitly rather than via the unknown-header path.

## Tests
New `ServerSideEncryptionTest` (9 cases): accept + echo on PUT/HEAD/GET/copy/complete; `ignore` accepts without echo; `reject` and `aws:kms`/SSE-C → 501; derived default = reject on transient and sse-s3 under `encrypted-blobstore`; `SseMode` parsing. Full `mvn verify -Pjdeps` is green (checkstyle, spotbugs, error-prone/NullAway, jdeps, 413 tests).

## Deliberately out of scope (later phases from #402)
- **Per-object persistence.** The echo is currently gated on the configured/derived mode rather than a persisted per-object marker. Happy to switch to an internal `s3proxy_*` metadata key (filtered from HEAD/list) if you prefer that shape.
- **SSE-C** for real (Phase 2), **bucket default encryption `?encryption`** (Phase 3), **SSE-KMS** (Phase 4).
- I added a focused JUnit test rather than enabling the existing `sse_s3` ceph tests in `run-s3-tests.sh` — glad to flip those on instead/as well.

Feedback welcome on the two design forks (mode-gated vs persisted marker; and whether the config knob is wanted at all vs a pure capability gate).

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,13 @@ S3Proxy has broad compatibility with the S3 API, however, it does not support:
 * hosting static websites
 * object lock, including legal hold and retention
 * object ownership controls
-* object server-side encryption, see [#402](https://github.com/gaul/s3proxy/issues/402)
+* object server-side encryption, see [#402](https://github.com/gaul/s3proxy/issues/402);
+  SSE-S3 (AES256) is accepted and echoed on PUT/HEAD/GET/copy/complete only when
+  objects are truly encrypted at rest -- a backend that always encrypts
+  (S3/Azure/GCS) or when `s3proxy.encrypted-blobstore` is enabled -- and rejected
+  otherwise, so the proxy never claims encryption it is not performing. Override
+  this default with `s3proxy.server-side-encryption` (`reject` | `ignore` |
+  `sse-s3`). SSE-KMS and SSE-C remain unsupported
 * object tagging
 * object versioning on backends other than aws-s3-sdk, see [#74](https://github.com/gaul/s3proxy/issues/74)
 * paginating ListParts with `part-number-marker`
--- a/src/main/java/org/gaul/s3proxy/AwsHttpHeaders.java
+++ b/src/main/java/org/gaul/s3proxy/AwsHttpHeaders.java
@@ -49,6 +49,14 @@ final class AwsHttpHeaders {
     static final String OBJECT_ATTRIBUTES = "x-amz-object-attributes";
     static final String REQUEST_ID = "x-amz-request-id";
     static final String SDK_CHECKSUM_ALGORITHM = "x-amz-sdk-checksum-algorithm";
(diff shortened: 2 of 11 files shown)


gaul:
CI failed due to the GitHub actions outage earlier today.

gavin-thompson-postman:
Closing this — server-side encryption has since been implemented upstream in 89948e02 ("Answer server-side encryption where the store can encrypt") and 9e200e9a ("Answer SSE-C on the stores that encrypt"), which supersede this draft and go further than it did.

The merged approach is a clean capability gate (`BlobStore.supportsServerSideEncryption()`) with real crypto (`blobstore/domain/Encryption`) covering both SSE-S3 and SSE-C on the stores that can encrypt — versus the config-driven SSE-S3-only shim here. I pulled master and verified: the `ServerSideEncryptionTest` suite passes (14/14) and a live SSE-S3 round-trip echoes `AES256` on PUT/HEAD/GET against an encrypting backend.

Thanks for the detailed design notes on #402 — they made it easy to see this was the better path. Closing as superseded.

PR #1131, opened 2026-08-07
Answer server-side encryption where the store can encrypt

gaul:
The blobstore interface has spoken the SDK's own request and response types
since the jclouds removal, so SSE-S3 and SSE-KMS need no new plumbing: the
put, copy and initiate-multipart handlers copy the
x-amz-server-side-encryption family onto requests they already build, the
aws-s3 store forwards them verbatim like everything else, and every response
reports what the backend answered rather than what was asked -- real S3
encrypts even what nobody asked it to, and only the backend knows which key.
SSE-C stays refused everywhere, because honoring a caller's key means
carrying it on every read and copy, not echoing it back on the write that
supplied it.  The MultipartUpload record grows the backend's creation answer
so the initiate response can carry it.

The transient store answers the family itself rather than forwarding it,
which is what the tests stand on: a store written to stand in for one that
encrypts is a second implementation of the thing under test, agreeing with
the frontend while telling nobody whether a store S3Proxy ships agrees too.
The two nio2 stores part over this on the line they already part over for
versioning.  What the transient store holds never rests anywhere a caller
could reach, its filesystem living and dying with the process, so an
object's encryption is the header family and nothing besides -- which is all
S3 lets a caller observe of SSE-S3 anyway.  The filesystem store writes to a
real disk, where reporting AES256 over plaintext someone could go and read
would be a claim S3Proxy has no business making; it answers the family
NotImplemented as it did while these headers were unrecognized, and reports
no encryption at all rather than the default, the default being that same
claim made unprompted.

An object's encryption rides in user.sse-* xattrs, and an object encrypted
the way every unasked-for object is carries none of them: an absent
algorithm reads back as the default, so the common case costs no attribute.
A multipart upload keeps its own on the stub blob, since the carriers built
for the upload's later requests name only the bucket and key -- by the time
a part is uploaded or the object completed, what the create asked for exists
nowhere else.  The part copy is emulated for a store with no server-side
copy of its own, reading the source back through the ordinary part upload;
the response built around that write dropped the encryption the write had
just reported, which the s3-tests suite caught where the tests here did not.

Both lanes whose store encrypts run the SSE-S3 tests now -- LocalStack,
which the aws-s3 backend forwards to, and the transient store -- and
twenty-nine pass on each.  Admitting the tag takes more than dropping it,
since eighteen of the twenty-one tests carry the encryption tag as well and
the expression has to let sse_s3 back through that exclusion rather than
lift it.  What that tag covers beyond SSE-S3 is SSE-C; bucket_encryption
goes out with it, those tests asking a bucket to encrypt by default, which
is the unimplemented ?encryption subresource rather than the header family.
The thirteen that do not pass are marked: ten copies with SSE-C on one end
and the three bucket encryption configuration tests.  kms_keyid2 becomes the
key's ARN because that is what a response carries -- S3 and LocalStack alike
answer with the ARN rather than the bare id they were handed, and the proxy
was checked against LocalStack directly, with and without it in the way,
returning the two byte for byte.

See #402.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ S3Proxy has broad compatibility with the S3 API, however, it does not support:
 * hosting static websites
 * object lock, including legal hold and retention
 * object ownership controls
-* object server-side encryption, see [#402](https://github.com/gaul/s3proxy/issues/402)
+* object server-side encryption on backends other than aws-s3-sdk, and SSE-C on any backend, see [#402](https://github.com/gaul/s3proxy/issues/402)
 * object tagging
 * object versioning on backends other than aws-s3-sdk, see [#74](https://github.com/gaul/s3proxy/issues/74)
 * paginating ListParts with `part-number-marker`
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ S3Proxy has broad compatibility with the S3 API, however, it does not support:
 * hosting static websites
 * object lock, including legal hold and retention
 * object ownership controls
-* object server-side encryption, see [#402](https://github.com/gaul/s3proxy/issues/402)
+* object server-side encryption with a customer-provided key (SSE-C), see [#402](https://github.com/gaul/s3proxy/issues/402)
 * object tagging
 * paginating ListParts with `part-number-marker`
 * public access block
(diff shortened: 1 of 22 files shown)


gaul:
@gavin-thompson-postman I want to propose this as an alternative first step towards SSE support instead of #1130.  While the diff is large due to tests, the main code just passes through S3 SSE headers to the S3 backend and raises 501 everywhere else.  I prefer this limited approach because it defers more difficult questions about what the filesystem format should be, how to interact with Azure's encryption, and how this might interact with the encryption middleware.  Actually I think the transient backend could add SSE (and versioning) and raise 501 for the filesystem but will defer that for another commit.  It would be good to hear more about your use case and whether this helps.

gaul:
I expanded the scope to pass through SSE to both s3 and transient.  Merging since this addresses s3fs's use case for SSE.

PR #1133, opened 2026-08-10
Answer SSE-S3 on the azureblob backend

gavin-thompson-postman:
Opening as a draft: this extends server-side encryption to the **azureblob** backend, which `89948e02` / `9e200e9a` left opted out.

## What
`AzureBlobStore` now:
- answers `supportsServerSideEncryption()`;
- reports **SSE-S3 (`AES256`)** on PUT / HEAD / GET / copy / complete — for every object, the way S3 reports default bucket encryption;
- **refuses SSE-C and SSE-KMS with 501** (`refuseUnsupportedSse`), rather than silently storing data under a key it can't honor.

## Why this is truthful for Azure
Azure Blob **always encrypts at rest** with Microsoft-managed keys and cannot be turned off, so acknowledging SSE-S3 is a true statement about the object — the same shape as your `testReportsDefaultEncryption` on the transient store ("what S3 applies to an object whose write asked for nothing"). This treats Azure's at-rest encryption as the store's encryption, which is how "answer SSE where the store can encrypt" applies here.

The line I did **not** cross: SSE-C and SSE-KMS imply key handling Azure's path here does not perform, so they're refused (501) — never accepted-and-ignored.

## Tests
New `ServerSideEncryptionAzuriteTest`, wired into the existing `azuriteTests` CI job:
- SSE-S3 accepted + echoed on PUT/HEAD/GET;
- default encryption reported for an unasked-for write;
- SSE-C refused (501);
- SSE-KMS refused (501).

It's gated with `assumeTrue` on the azureblob backend, so it runs under `-Ds3proxy.test.conf=s3proxy-azurite.conf` and **skips** on backends that implement SSE-C. `mvn verify -Pjdeps` is green (479 tests); verified locally against Azurite (4/4) and against a real Azure Storage account.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -222,7 +222,7 @@ jobs:
             'http://127.0.0.1:10000/devstoreaccount1?comp=list'
       - name: Maven Test with Azurite
         run: |
-          mvn test -Ds3proxy.test.conf=s3proxy-azurite.conf -Dtest=AwsSdkTest,AwsSdkAnonymousTest,AwsSdkServicePathTest,AwsSdkVirtualHostTest,CrossOriginResourceSharingAllowAllResponseTest,CrossOriginResourceSharingResponseTest,EncryptedBlobStoreTest,MultipartUploadKeyCharacterTest
+          mvn test -Ds3proxy.test.conf=s3proxy-azurite.conf -Dtest=AwsSdkTest,AwsSdkAnonymousTest,AwsSdkServicePathTest,AwsSdkVirtualHostTest,CrossOriginResourceSharingAllowAllResponseTest,CrossOriginResourceSharingResponseTest,EncryptedBlobStoreTest,MultipartUploadKeyCharacterTest,ServerSideEncryptionAzuriteTest
       - uses: actions/upload-artifact@v7
         if: always()
         with:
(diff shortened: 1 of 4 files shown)


gaul:
The azuriteTests failure was not encryption-related. `test_list_buckets_paginated` requires an account that owns no buckets, and the new `ServerSideEncryptionAzuriteTest` was leaking containers into the s3-tests step, which runs against the same Azurite instance as the Maven suites: each test created a container that `tearDown` never deleted, and the s3-tests cleanup only removes buckets matching its own prefixes, so the first pagination assertion saw a stray `container-1214152011`.

I pushed 40d7289 to this branch, which deletes the container in `tearDown` following the pattern of the other suites (`BlobStore.deleteContainer` empties the container first, covering the tests that leave an object behind). Locally the suite passes 4/4 against Azurite and the account lists no containers afterward.

gaul:
Initial CI failed and the provider reported several incorrect attributes.  Merged with multiple fixes.  References #402.  Thank you for your contribution @gavin-thompson-postman!

gavin-thompson-postman:
@gaul Thanks for taking this to completion.

I had raised this as a draft yesterday as I wanted you (and others) to see my intention, but I needed some time to resolve the failing CI tests. Thanks for taking care of that part.

Issue #1137, opened 2026-08-11
Object versioning for the azureblob backend


gaul:
Remainder of #74 for the `azureblob` backend, which reports `supportsVersioning() == false` and refuses a named `versionId` outright on reads, copies and multipart copies (`AzureBlobStore.java:422`, `:784`, `:1049`, `:1480`).

Azure does version blobs, and the data-plane vocabulary lines up: a write answers `x-ms-version-id`, `getVersionClient(versionId)` addresses one version, and `BlobListDetails.setRetrieveVersions(true)` lists them.  What does not line up is everything around that.

- **The switch is account-level and outside the API S3Proxy speaks.**  Blob versioning is turned on for an entire storage account through the management plane, not per container and not through the data-plane SDK.  `PUT ?versioning Enabled` on one bucket therefore has nowhere to go: the store can report only what the account already does, or keep a per-container flag of its own and refuse the request when the account has versioning off.  Settling that is most of this issue; the rest is mechanical.
- **Suspended has no analog** either, the same way it has none on GCS (#1136).
- **No delete markers.**  Deleting a blob under versioning deletes the current version, and nothing stands in the listing saying the key was deleted -- Azure's soft delete is a differently shaped feature, not this one.  Emulating a marker, a zero-byte version carrying metadata, is the same choice #1136 faces.
- **Snapshots are the older analog and a worse one.**  They are addressable and predate versioning, but nothing creates one on overwrite, so the store would have to snapshot before every write: two round trips that no lease makes atomic.

The SPI to fill in is the same list as every other backend -- `getContainerVersioning`/`setContainerVersioning`, `listVersions`, the `versionId` arms of `getBlob`, `blobMetadata`, `copyBlob`, `removeBlob`, `getBlobAccess` and `setBlobAccess`, and the version ids reported by `putBlob`, `copyBlob` and `completeMultipartUpload` -- with `aws-s3` (70e7a79) as the model for a store whose service already versions.

The testing story is the hard part, as with #1135: Azurite lists Blob Versions among the features it does not implement, so the `s3proxy-azurite.conf` lane cannot cover this until that changes upstream, leaving a real storage account as the alternative.  Note too that `test_object_copy_versioning_multipart_upload` and `test_multipart_copy_versioned` already carry `fails_on_s3proxy_azureblob` marks while that lane deselects the `versioning` tag: they are not exercised today, and the marks want re-verifying whenever the tag runs.

PR #1139, opened 2026-08-11
Version objects on azureblob

gavin-thompson-postman:
Maps S3 object versioning onto Azure's native blob versioning for the `azureblob` backend. Partially addresses #1137.

**This is phase 1 of 3, and is written to be reviewable and mergeable on its own** — it stands up the full version-*use* surface, and each later phase adds an independent capability without reworking this one:
- **Phase 1 (this PR)** — native version mapping: version ids, read/head/delete/copy by version, `ListObjectVersions`.
- **Phase 2** — delete-marker emulation (null-version delete inserts a marker, undelete, markers in `ListObjectVersions`).
- **Phase 3** — suspended-state semantics (null version, suspended delete markers).

## Covered
- Version ids on `PutObject`, `CopyObject` and `CompleteMultipartUpload`
- `GET` / `HEAD` / `DELETE` / copy by version id
- `ListObjectVersions` (versions, `IsLatest`, prefix, pagination)

## Opt-in
Azure enables versioning per storage account on the management plane (ARM); the data-plane SDK can neither set nor read the setting. So it is opt-in through `s3proxy.azureblob.versioning=true`, by which the operator asserts the account already has versioning on. Left unset, every versioning request answers `NotImplemented` exactly as before — no change for existing configs.

## Azure quirks handled
- Deleting the current ("root") version by id is refused by Azure (`OperationNotAllowedOnRootBlob`); handled by deleting the base blob first, then the demoted version.
- A copy-by-version source URL already carries `?versionid=`, so the SAS joins with `&` (a bare `?` yields `CannotVerifyCopySource`).
- A malformed/unknown version id surfaces as `400` rather than `500`.

## Deferred to later phases (not faked)
Azure has no analog, so these stay unimplemented rather than emulated for now:
- Delete markers — a null-version delete drops the current version and retains the prior ones, but writes no marker, and there is no undelete. (phase 2)
- Per-bucket suspend — `setContainerVersioning(SUSPENDED)` returns `501`. (phase 3)

## Testing
Azurite does not emulate blob versioning today, so there is no Azurite CI lane for this (unlike the SSE work). This PR adds a unit test for the opt-in gating that runs without a backend. The version-use paths were validated end-to-end against a real Azure storage account (with account versioning enabled) using an external S3 versioning conformance suite; the deferred items above are the only expected gaps. `mvn verify -Pjdeps` is green.

I am tracking [Azure/Azurite#665](https://github.com/Azure/Azurite/issues/665), which has had some recent activity — if/when Azurite gains blob-versioning support I will wire an Azurite-backed CI lane for this the way the SSE work has one. If that issue stalls, I will attempt to contribute the versioning support to Azurite myself so this can be covered in CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ Some limitations depend on the storage backend:
 | `max-keys=0` not honored | `azureblob` |
 | `UploadPartCopy` streams the data through S3Proxy instead of copying it on the backend | `filesystem`, `transient`, `openstack-swift`, `sftp` |
 | conditional PUT refuses `If-Match`, honoring only `If-None-Match: *` | `openstack-swift` |
-| no object versioning, see [#1137](https://github.com/gaul/s3proxy/issues/1137) | `azureblob` |
+| object versioning is opt-in and needs the account versioned out of band; no delete markers or per-bucket suspend, see [#1137](https://github.com/gaul/s3proxy/issues/1137) | `azureblob` |
 | no object versioning, see [#1138](https://github.com/gaul/s3proxy/issues/1138) | `filesystem` |
 | no object versioning, for want of a service analog | `openstack-swift`, `sftp` |
 | no server-side encryption, see [#1134](https://github.com/gaul/s3proxy/issues/1134) | `google-cloud-storage` |
@@ -183,7 +183,12 @@ simply on or off, it refuses the Suspended state.  `transient` implements
 versioning itself, keeping every version but the current one as a hidden
 object.  `filesystem` shares that implementation and declines it: its versions
 would outlive the process, settling a layout on disk that S3Proxy would owe
-its users from then on.
+its users from then on.  `azureblob` maps onto Azure's own blob versioning,
+which is turned on per storage account on the management plane (ARM) and
+cannot be read or set through the data-plane SDK -- so it is opt-in through
+`s3proxy.azureblob.versioning=true`, by which the operator asserts the account
+has versioning on.  What Azure has no analog for it does not emulate: there
+are no delete markers, and versioning cannot be suspended per bucket.
 
 Server-side encryption divides `aws-s3` and `transient` from most of the
 rest, and divides the two nio2 stores on the same line.  `aws-s3` forwards the header families --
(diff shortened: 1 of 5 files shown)


gaul:
It would be better to start with a PR for Azurite which will automate testing, particularly via s3-tests.  I might consider using an fork to test since I also want the nextMarker functionality.  This is not a hard blocker but I recommend pursuing this approach.

gavin-thompson-postman:
@gaul Thanks for the input. 

I have added some unit tests that give partial coverage to the new code in the absence of Azurite support. I agree with you that ideally the Azurite PR would happen first, but I'm a bit worried about the timelines from our side.

Would you be open to this being merged with the additional unit tests for now?

gavin-thompson-postman:
@gaul I raised https://github.com/Azure/Azurite/pull/2734 to try to get the Azurite support in first.

gavin-thompson-postman:
@gaul One of the MSFT team actually raised their own PR for this https://github.com/Azure/Azurite/pull/2735. If that one gets merged I will update the tests to run against Azurite.

I notice Azurite doesn't get released that often, so to get this working in the CI for s3proxy I guess a fork (as you mentioned) or specific sha of Azurite would need to be pulled in temporarily until Azurite releases.

gavin-thompson-postman:
@gaul - Would you be comfortable with this being merged in its current state with the additional unit tests that I put in place?

https://github.com/Azure/Azurite/pull/2735 is still pending review unfortunately so I can't plug in Azurite for these flows in the CI.

I'd be happy to revisit this make sure the CI does make full use of Azurite if that PR does get merged. 

Thanks!

gaul:
> @gaul - Would you be comfortable with this being merged in its current state with the additional unit tests that I put in place?

I skipped this PR for 4.1.0.  Azurite has shown some movement on the original implementation and your proposed PR so let's wait and see.  I am willing to pin against a remote hash once they review and merge one of them so we don't have to wait for a versioned release.

I don't believe that AI-generated unit tests are convincing.  AIs regularly hallucinate and it is essential to ground them in reality, in this case using s3-tests against the live cloud service or using emulator.  It is theoretically possible to set up a test against live Azure but every other S3Proxy feature has CI coverage through emulators so I don't want to make an exception unless I have to.

Issue #74, opened 2015-07-24
Object versioning


andrewgaul:
S3Proxy does not currently support object versioning, triggering many s3-tests failures:

```
ERROR: s3tests.functional.test_s3.test_versioning_bucket_create_suspend
ERROR: s3tests.functional.test_s3.test_versioning_obj_create_read_remove
ERROR: s3tests.functional.test_s3.test_versioning_obj_create_read_remove_head
ERROR: s3tests.functional.test_s3.test_versioning_obj_suspend_versions
ERROR: s3tests.functional.test_s3.test_versioning_obj_suspend_versions_simple
ERROR: s3tests.functional.test_s3.test_versioning_obj_create_versions_remove_all
ERROR: s3tests.functional.test_s3.test_versioning_obj_create_overwrite_multipart
ERROR: s3tests.functional.test_s3.test_versioning_obj_list_marker
ERROR: s3tests.functional.test_s3.test_versioning_copy_obj_version
ERROR: s3tests.functional.test_s3.test_versioning_multi_object_delete
ERROR: s3tests.functional.test_s3.test_versioning_multi_object_delete_with_marker
ERROR: s3tests.functional.test_s3.test_versioning_multi_object_delete_with_marker_create
ERROR: s3tests.functional.test_s3.test_versioned_object_acl
ERROR: s3tests.functional.test_s3.test_versioned_concurrent_object_create_concurrent_remove
ERROR: s3tests.functional.test_s3.test_versioned_concurrent_object_create_and_remove
```

Implementing this requires upstream jclouds work tracked by [JCLOUDS-895](https://issues.apache.org/jira/browse/JCLOUDS-895).

jeffcollaboro:
Hi Andrew, just bumping this. I see you have done quite a bit of work in the JCLouds upstream but it seems to have lost momentum over a few years of being blocked by issues. Wondering your thoughts on dusting this off.

gaul:
This is a large task and I have no plans to work on this.  I don't think jclouds is a good base for the future, instead S3Proxy should use vendor SDKs as suggested in #606 which implements the jclouds `BlobStore` interface.  Then S3Proxy can create and implement an `ExtendedBlobStore` interface that calls into the versioned APIs when available.  Versioning specifically creates some issues since we need the version number in the object metadata.

gaul:
[...]

PR #1140, opened 2026-08-17
Object versioning for the google-cloud-storage backend

### LINKED ISSUE #1136: Object versioning for the google-cloud-storage backend
Remainder of #74 for the `google-cloud-storage` backend, which reports `supportsVersioning() == false`, so every versioning request is 501 apart from `GET ?versioning` -- answered everywhere as a 200 empty configuration, S3's "never versioned".

Two backends version objects today and neither pattern carries over unchanged: `aws-s3` passes the requests through to a service that already speaks S3 (70e7a79) and `transient` implements the semantics itself over hidden sidecar files (dad8a72).  GCS wants a third shape, a translation, because its analog is close without being congruent.

- **Generations are the version ids.**  `BlobId.of(bucket, name, generation)` addresses one version, `BlobListOption.versions(true)` lists the noncurrent ones, and every write reports the generation it created.  The store already pins generations to make conditional writes atomic (`GCloudBlobStore.java:570-598`), so naming a version needs nothing new -- only exposing it.
- **Enabling versioning is a data-plane setting.**  `BucketInfo.setVersioningEnabled` rides the ordinary JSON API, so `PUT ?versioning` maps straight onto it, unlike the azureblob case where the switch sits outside the API S3Proxy speaks.
- **Suspended has no analog.**  GCS versioning is on or off, where S3's third state keeps existing versions addressable while writes replace the `null` version and report no version id.  Turning GCS versioning off preserves the noncurrent generations, which is half of it; the null-version rule would have to be emulated, or the state refused.
- **Delete markers are the real gap.**  GCS has no such object: deleting a live generation under versioning archives it and the next `GET` is a plain 404.  S3 wants the `DELETE` to answer `x-amz-delete-marker: true` with a version id, the marker to appear in `ListObjectVersions`, a `GET` of it to be 405, and deleting the marker to promote the object back.  Either emulate one -- a zero-byte generation carrying custom metadata, the trick `transient` plays with its `.s3proxy-versions` entries -- or accept the divergence and let a delete be permanent, which forfeits a good part of the versioning tag.

The SPI to fill in is `getContainerVersioning`/`setContainerVersioning`, `listVersions`, and the `versionId` arms of `getBlob`, `blobMetadata`, `copyBlob`, `removeBlob`, `getBlobAccess` and `setBlobAccess`, plus the version ids reported by `putBlob`, `copyBlob` and `completeMultipartUpload`.  `BlobStore` documents each contract, including the `noSuchKeyDeleteMarker` and `noSuchVersion` exceptions the handler turns into S3's answers, and the per-version ACL overloads that 4e6fcdb added.

Unlike #1134 the testing story is the easy part: fake-gcs-server keeps `activeObjects` and `archivedObjects` with generations and answers `versions=true` listings in its memory backend, which is the one CI runs (`-backend memory`), so `run-s3-tests.sh` can stop deselecting the `versioning` tag on the `s3proxy-fake-gcs-server.conf` lane and let the suite say how far the translation got.

gaul:
Fixes #1136, the remainder of #74 for this backend.  Where aws-s3 passes versioning through and transient implements it, GCS wants a translation: generations are the version ids -- BlobId.of(bucket, name, generation) addresses one, every write reports the one it created -- and PUT ?versioning rides BucketInfo.setVersioningEnabled over the ordinary JSON API.

What GCS lacks is emulated or refused:

- Delete markers become zero-byte generations carrying s3proxy_delete_marker metadata.  Reads, listings and copies know them: a current marker answers 404 with x-amz-delete-marker, a read naming one answers 405, ListObjectVersions renders them as DeleteMarker entries, and unversioned listings hide their keys.

- Deleting a named version removes that generation outright, which on the newest promotes whatever it hid -- resolved on the read side: when no live object answers, the newest remaining generation does.

- Suspended is refused with NotImplemented.  Turning GCS versioning off keeps noncurrent generations but replaces the live object permanently on the next write, so the null-version rule cannot be told from data loss.

Whether a container versions decides if writes report version ids and deletes leave markers; S3 asks per operation while GCS answers only through bucket metadata, so a small cache holds the answer briefly. Multipart bookkeeping -- parts, intermediate composes, the stub -- is now removed generation by generation, since a plain delete on a versioned bucket would archive it forever.  Completion also gains native conditional publishes: a named ETag resolves to the matching generation and pins the compose, as a plain put already did, so the NATIVE_CONDITIONAL_COMPLETE quirk admits this backend.

TransientVersioningTest becomes VersionedBlobStoreTest and runs on the fake-gcs-server lane as well; stores that refuse Suspended skip those tests.  The s3-tests lane stops deselecting the versioning tag: 25 tests pass and 11 carry markers -- the Suspended and null-version families, and the recreate-a-bucket 409 the nio2 lane also marks. The lane needs fake-gcs-server built from the versioning-fixes branch, which teaches it generation-addressed deletes, fresh compose generations and per-generation ACLs; CI builds that branch until the fixes land upstream.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -334,7 +334,17 @@ jobs:
           mvn package -DskipTests
 
       - name: Install fake-gcs-server
-        run: go install github.com/fsouza/fake-gcs-server@v1.55.0
+        # Built from a branch until the fixes the versioning translation
+        # leans on -- deleting a named generation, compose minting a new
+        # generation, per-generation object ACLs -- are released upstream;
+        # then this goes back to go install fsouza/fake-gcs-server@vNEXT.
+        run: |
+          git clone --depth 1 --branch versioning-fixes \
+              https://github.com/gaul/fake-gcs-server.git \
+              "$RUNNER_TEMP/fake-gcs-server"
+          mkdir -p "$HOME/go/bin"
+          cd "$RUNNER_TEMP/fake-gcs-server"
+          go build -o "$HOME/go/bin/fake-gcs-server" .
       - name: Start fake-gcs-server
         run: |
           $HOME/go/bin/fake-gcs-server -backend memory -scheme http -host 127.0.0.1 &
@@ -342,7 +352,7 @@ jobs:
             http://127.0.0.1:4443/storage/v1/b
       - name: Maven Test with fake-gcs-server
         run: |
-          STORAGE_EMULATOR_HOST=http://localhost:4443 mvn test -Ds3proxy.test.conf=s3proxy-fake-gcs-server.conf -Dtest=AwsSdkTest,AwsSdkAnonymousTest,AwsSdkServicePathTest,AwsSdkVirtualHostTest,CrossOriginResourceSharingAllowAllResponseTest,CrossOriginResourceSharingResponseTest,EncryptedBlobStoreTest
+          STORAGE_EMULATOR_HOST=http://localhost:4443 mvn test -Ds3proxy.test.conf=s3proxy-fake-gcs-server.conf -Dtest=AwsSdkTest,AwsSdkAnonymousTest,AwsSdkServicePathTest,AwsSdkVirtualHostTest,CrossOriginResourceSharingAllowAllResponseTest,CrossOriginResourceSharingResponseTest,EncryptedBlobStoreTest,VersionedBlobStoreTest
       - uses: actions/upload-artifact@v7
         if: always()
         with:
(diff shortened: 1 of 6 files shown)


gaul:
Blocked on fsouza/fake-gcs-server#2322, fsouza/fake-gcs-server#2323, fsouza/fake-gcs-server#2324, and one more unsubmitted PR.

gaul:
Pushed as 6ed66c737fd8f4a707bb44298af0af264cf47552.

